PR #33, opened 2024-07-25
Fix a few compile errors

koddsson:
I figured I'd start small so that I'm sure I'm not doing something really wrong. I'll make bigger changes as I go along.


crates/hdx_parser/src/traits/mod.rs
```diff
@@ -154,6 +154,7 @@ pub trait RangedMediaFeature<'a>: Sized {
 	fn new(left: (Comparison, Self::Type), right: Option<(Comparison, Self::Type)>, legacy: bool) -> Self;
 
 	fn parse_ranged_media_feature(name: Atom, parser: &mut Parser<'a>) -> Result<Self> {
+		let checkpoint = parser.checkpoint();
```
koddsson, line 157: Is this correct? Set a checkpoint, then call `.next()` and of we don't get the token we want, then we rewind to that checkpoint?

keithamus: yep that's exactly it!

keithamus (approved):
Nice work!

PR #36, opened 2024-07-25
try a method for matching delimiters

koddsson:
How does this look? Is this correct? Should I peek and store that in a variable for performance? Or readability?

### DIFF
--- a/crates/hdx_parser/src/comparison.rs
+++ b/crates/hdx_parser/src/comparison.rs
@@ -18,18 +18,16 @@ impl<'a> Parse<'a> for Comparison {
 		Ok(if match_token_kind_and_char!(next, Kind::Delim, '=') {
 			Comparison::Equal
 		} else if match_token_kind_and_char!(next, Kind::Delim, '>') {
-			if let (Kind::Delim, Some('=')) =
-				(parser.peek_with(Include::Whitespace).kind(), parser.peek_with(Include::Whitespace).char())
-			{
+			let peek = parser.peek_with(Include::Whitespace);
+			if match_token_kind_and_char!(peek, Kind::Delim, '=') {
 				parser.advance_with(Include::Whitespace);
 				Comparison::GreaterThanEqual
 			} else {
 				Comparison::GreaterThan
 			}
 		} else if match_token_kind_and_char!(next, Kind::Delim, '<') {
-			if let (Kind::Delim, Some('=')) =
-				(parser.peek_with(Include::Whitespace).kind(), parser.peek_with(Include::Whitespace).char())
-			{
+			let peek = parser.peek_with(Include::Whitespace);
+			if match_token_kind_and_char!(peek, Kind::Delim, '=') {
 				parser.advance_with(Include::Whitespace);
 				Comparison::LessThanEqual
 			} else {
--- a/crates/hdx_parser/src/comparison.rs
+++ b/crates/hdx_parser/src/comparison.rs
@@ -1,5 +1,5 @@
 use crate::{unexpected, Parse, Parser, Result};
-use hdx_lexer::{Include, Token};
+use hdx_lexer::{Include, Kind};
 
 #[derive(Debug, PartialEq, Hash)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize), serde(tag = "type"))]
@@ -13,25 +13,29 @@ pub enum Comparison {
 
 impl<'a> Parse<'a> for Comparison {
 	fn parse(parser: &mut Parser<'a>) -> Result<Comparison> {
-		Ok(match parser.next() {
-			Token::Delim('=') => Comparison::Equal,
-			Token::Delim('>') => {
-				if let Token::Delim('=') = parser.peek_with(Include::Whitespace) {
+		Ok(match (parser.next().kind(), parser.next().char()) {
+			(Kind::Delim, Some('=')) => Comparison::Equal,
+			(Kind::Delim, Some('>')) => {
+				if let (Kind::Delim, Some('=')) =
+					(parser.peek_with(Include::Whitespace).kind(), parser.peek_with(Include::Whitespace).char())
+				{
 					parser.advance_with(Include::Whitespace);
 					Comparison::GreaterThanEqual
 				} else {
 					Comparison::GreaterThan
 				}
 			}
-			Token::Delim('<') => {
-				if let Token::Delim('=') = parser.peek_with(Include::Whitespace) {
+			(Kind::Delim, Some('<')) => {
+				if let (Kind::Delim, Some('=')) =
+					(parser.peek_with(Include::Whitespace).kind(), parser.peek_with(Include::Whitespace).char())
+				{
 					parser.advance_with(Include::Whitespace);
 					Comparison::LessThanEqual
 				} else {
 					Comparison::LessThan
 				}
 			}
-			token => unexpected!(parser, token),
+			(token, _) => unexpected!(parser, token),
 		})
 	}
 }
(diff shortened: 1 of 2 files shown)


crates/hdx_parser/src/comparison.rs
```diff
@@ -13,25 +13,29 @@ pub enum Comparison {
 
 impl<'a> Parse<'a> for Comparison {
 	fn parse(parser: &mut Parser<'a>) -> Result<Comparison> {
-		Ok(match parser.next() {
-			Token::Delim('=') => Comparison::Equal,
-			Token::Delim('>') => {
-				if let Token::Delim('=') = parser.peek_with(Include::Whitespace) {
+		Ok(match (parser.next().kind(), parser.next().char()) {
+			(Kind::Delim, Some('=')) => Comparison::Equal,
+			(Kind::Delim, Some('>')) => {
```
keithamus: I wonder if it would be nicer as a macro, kind of like we have for `match_ignore_case`?

koddsson: A challenge, eh? I'll give it a go 👯

crates/hdx_parser/src/comparison.rs
```diff
@@ -13,25 +13,29 @@ pub enum Comparison {
 
 impl<'a> Parse<'a> for Comparison {
 	fn parse(parser: &mut Parser<'a>) -> Result<Comparison> {
-		Ok(match parser.next() {
-			Token::Delim('=') => Comparison::Equal,
-			Token::Delim('>') => {
-				if let Token::Delim('=') = parser.peek_with(Include::Whitespace) {
+		Ok(match (parser.next().kind(), parser.next().char()) {
```
keithamus: IIRC `next()` advances the position so this wouldn't work, you might need something like this instead:

```suggestion
    let token = parser.next();
		Ok(match (token.kind(), token.char()) {
```

koddsson:
This _feels_ like this might be wrong but just let me know.

keithamus:
Want to try building a more complex macro using match arms? I’m sure we have something similar in the existing macros.

koddsson:
> Want to try building a more complex macro using match arms? I’m sure we have something similar in the existing macros.

Sure! I'm not actually sure what that entails here.

koddsson:
> > Want to try building a more complex macro using match arms? I’m sure we have something similar in the existing macros.
> 
> Sure! I'm not actually sure what that entails here.

Is it creating a macro that can be matched? Something that would enable this?

```rust
match token_kind_and_char!(token) {
  (Kind::Delim, '=') => todo!()
} 
```

keithamus:
> > > Want to try building a more complex macro using match arms? I’m sure we have something similar in the existing macros.
> > 
> > 
> > Sure! I'm not actually sure what that entails here.
> 
> Is it creating a macro that can be matched? Something that would enable this?
> 
> ```rust
> match token_kind_and_char!(token) {
>   (Kind::Delim, '=') => todo!()
> } 
> ```

Generally I'm trying to work towards a set of macros that take the shape of this:

```rust
expect_some_type!(parser.next());
expect_some_type!(parser.next(), some, details, about, the, type);

if match_some_type!(parser.next(), some, details, about, the, type) {
  // ...
}

match_some_type!{ parser, some, details, :
 foo => { /* ... */ }
 bar => { /* ... */ }
 baz => { /* ... */ }
}
```

In this way, we should write a `match_delim!` macro which allows those two types. So I'd expect `match_delim!(parser.next(), '-')` to be the equivalent of `matches!(parser.next(), Token::Delim('-'))`. But I'd expect the following two blocks to also be equivalent:

```rust
match_delim!{ parser.next() :
  '=' => true,
  '>' => true,
  '<' => true,
  _ => false,
}

let token = parser.next();
match token.kind():
  Kind::Delim => match token.char() {
    '=' => true,
    '>' => true,
    '<' => true,
    _ => false,
  },
  _ => false,
}
```

You can see I have `expect_ignore_case!`, and `match_ignore_case!` (you can see this used [in an if here](https://github.com/keithamus/hdx/blob/main/crates/hdx_ast/src/css/rules/media/mod.rs#L169) and [as a match block here](https://github.com/keithamus/hdx/blob/rewrite-lexer/crates/hdx_ast/src/css/types/color/mod.rs#L198-L202)). which represent this pattern. We have `expect_delim!`, so now we just need `match_delim!`. 

[`match_ignore_case` is defined here](https://github.com/keithamus/hdx/blob/rewrite-lexer/crates/hdx_parser/src/macros.rs#L259-L284) and it _looks_ very complicated, but I think you could copy-paste this into a `match_delim` and tweak it. You won't need the `Token::$tokenty` bits as Delim is of course only one type.

crates/hdx_parser/src/comparison.rs
```diff
@@ -13,25 +14,28 @@ pub enum Comparison {
...
-			Token::Delim('=') => Comparison::Equal,
-			Token::Delim('>') => {
-				if let Token::Delim('=') = parser.peek_with(Include::Whitespace) {
-					parser.advance_with(Include::Whitespace);
-					Comparison::GreaterThanEqual
-				} else {
-					Comparison::GreaterThan
+		let next = parser.next();
+		Ok(match_delim! {parser.next() :
+		  '=' => Comparison::Equal,
```
koddsson: I feel like the formatting is all messed up here?

crates/hdx_parser/src/macros.rs
```diff
@@ -282,3 +282,17 @@ macro_rules! match_ignore_case {
...
+           $(
+              $pattern:pat $(if $guard:expr)?  => $then: expr
+           ),+
+           $(,)?
+       ) => {
+		match $parser.$method($($args)*).char() {
+		    $(Some($pattern) $( if $guard )? => $then,)+
+		}
+	};
+}
```
koddsson, line 298: @keithamus: something like this? I'm noticing now that I don't have the `.kind()` check in here now. I still need that right? Should I do the match here for a tuple like I was doing earlier and always make sure that `.kind() == Kind::Delim`?

keithamus: Ideally it should check `.kind()` because `.char()` returns values for the other single character tokens like `RightParen`. In this particular case it's okay because we have a fallthrough to `unexpected!` but we should refine this later.

crates/hdx_parser/src/comparison.rs
keithamus: Formatting is okay here.

crates/hdx_parser/src/comparison.rs
```diff
@@ -13,25 +14,28 @@ pub enum Comparison {
...
-		Ok(match parser.next() {
-			Token::Delim('=') => Comparison::Equal,
-			Token::Delim('>') => {
-				if let Token::Delim('=') = parser.peek_with(Include::Whitespace) {
-					parser.advance_with(Include::Whitespace);
-					Comparison::GreaterThanEqual
-				} else {
-					Comparison::GreaterThan
+		let next = parser.next();
+		Ok(match_delim! {parser.next() :
```
keithamus: This still forwards the parser twice. 
```suggestion
		Ok(match_delim! {parser.next() :
```

crates/hdx_parser/src/comparison.rs
```diff
@@ -13,25 +14,28 @@ pub enum Comparison {
...
+					'=' => {
+								parser.advance_with(Include::Whitespace);
+								Comparison::LessThanEqual
+					},
+					_ => Comparison::LessThan
 				}
-			}
-			token => unexpected!(parser, token),
+			},
+			_ =>  unexpected!(parser, next)
```
keithamus: Given above you'll need to:

```suggestion
			token =>  unexpected!(parser, token)
```

keithamus (approved):
Nice work!

PR #37, opened 2024-07-28
fix more compile errors in parsing token kind

koddsson:
I messed around some more. Not 100% sure this is the correct way to do this still.

### DIFF
--- a/crates/hdx_parser/src/traits/declarations.rs
+++ b/crates/hdx_parser/src/traits/declarations.rs
@@ -1,7 +1,7 @@
 use hdx_atom::{atom, Atom};
-use hdx_lexer::{Include, Kind, Token};
+use hdx_lexer::{Include, Kind};
 
-use crate::{discard, expect, expect_ignore_case, match_ignore_case, parser::Parser, unexpected, Result};
+use crate::{discard, expect, expect_ignore_case, parser::Parser, unexpected, Result};
 
 use super::Parse;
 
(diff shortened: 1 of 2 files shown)


crates/hdx_parser/src/traits/declarations.rs
```diff
@@ -1,26 +1,31 @@
...
 
 	fn parse_name(parser: &mut Parser<'a>) -> Result<Atom> {
-		match parser.next().clone() {
-			Token::Ident(atom) => {
+		let token = parser.next();
+		match token.kind() {
+			Kind::Ident => {
 				expect!(parser.next(), Kind::Colon);
-				Ok(atom.to_ascii_lowercase())
+				Ok(parser.parse_atom_lower(token))
```
koddsson, line 16: Is this the correct way to get a atom from a token ident now?

crates/hdx_parser/src/traits/declarations.rs
```diff
@@ -1,26 +1,31 @@
...
 
 	fn parse_important(parser: &mut Parser<'a>) -> Result<bool> {
-		if matches!(parser.peek(), Token::Delim('!'))
-			&& match_ignore_case!(parser.peek_n(2), Token::Ident(atom!("important")))
+		let peeked = parser.peek();
+		let nexted_peek = parser.peek_n(2);
+		if matches!(peeked.kind(), Kind::Delim)
+			&& matches!(peeked.char(), Some('!'))
+			&& matches!(nexted_peek.kind(), Kind::Ident)
+			&& matches!(parser.parse_atom_lower(*nexted_peek), atom!("important"))
```
koddsson, line 28: Is this ok? Should this have a macro or is there a better way?

crates/hdx_parser/src/traits/declarations.rs
keithamus: Yep! The new Lexer avoids emitting or assigning Atoms until you explicitly call something like `parse_atom()`/`parse_atom_lower()`. This lets us skip a bunch of work+memory allocations.

crates/hdx_parser/src/traits/declarations.rs
keithamus: So you should be able to do something like:

```suggestion
		if match_delim!(parser.peek(), Kind::Delim('!'))
		  && match_ignore_case!(parser.peek_n(2), Kind::Ident(atom!("important"))
```

But I think for this to work we'd need to do some work to both the match macros used.

keithamus (approved):
This looks good! I'm going to merge this and follow up with some refinement.

PR #61, opened 2024-12-10
(ast) add color_hdr support

keithamus:
This adds supports for the [CSS color-hdr](https://drafts.csswg.org/css-color-hdr-1/), by commenting out the `DynamicRangeLimit` enum and adding a new `DynamicRangeLimitMix` struct in the `crates/hdx_ast/src/css/values/color_hdr/types.rs` file. 

DynamicRangeLimitMix is a struct that represents the `dynamic-range-limit-mix()` function in CSS.

### DIFF
--- a/crates/hdx_ast/src/css/values/color_hdr/impls.rs
+++ b/crates/hdx_ast/src/css/values/color_hdr/impls.rs
@@ -1,1 +1,20 @@
 pub(crate) use crate::traits::StyleValue;
+pub(crate) use hdx_proc_macro::*;
+
+#[cfg(test)]
+mod tests {
+	use super::super::*;
+	use crate::test_helpers::*;
+
+	#[test]
+	fn size_test() {
+		assert_size!(DynamicRangeLimit, 56);
+	}
+
+	#[test]
+	fn test_writes() {
+		assert_parse!(DynamicRangeLimit, "high");
+		assert_parse!(DynamicRangeLimit, "dynamic-range-limit-mix(high 80%,standard 20%)");
+		assert_parse!(DynamicRangeLimit, "dynamic-range-limit-mix(high 8%,standard 2%)");
+	}
+}
--- a/crates/hdx_ast/src/css/values/color_hdr/mod.rs
+++ b/crates/hdx_ast/src/css/values/color_hdr/mod.rs
@@ -8,12 +8,12 @@ use impls::*;
  * CSS Color HDR Module Level 1
  */
 
-// // https://drafts.csswg.org/css-color-hdr-1/#dynamic-range-limit
-// #[value(" standard | high | constrained-high | <dynamic-range-limit-mix()> ")]
-// #[initial("high")]
-// #[applies_to("all elements")]
-// #[inherited("no")]
-// #[percentages("n/a")]
-// #[canonical_order("per grammar")]
-// #[animation_type("by dynamic-range-limit-mix()")]
-// pub enum DynamicRangeLimit {}
+// https://drafts.csswg.org/css-color-hdr-1/#dynamic-range-limit
+#[value(" standard | high | constrained-high | <dynamic-range-limit-mix()> ")]
+#[initial("high")]
+#[applies_to("all elements")]
+#[inherited("no")]
+#[percentages("n/a")]
+#[canonical_order("per grammar")]
+#[animation_type("by dynamic-range-limit-mix()")]
+pub enum DynamicRangeLimit<'a> {}
--- a/crates/hdx_ast/src/css/values/color_hdr/types.rs
+++ b/crates/hdx_ast/src/css/values/color_hdr/types.rs
@@ -1,2 +1,82 @@
-pub(crate) use crate::css::types::*;
-pub(crate) use crate::css::units::*;
+use bumpalo::collections::Vec;
+use hdx_lexer::Cursor;
+use hdx_parser::{diagnostics, CursorStream, Is, Parse, Parser, Peek, Result as ParserResult, ToCursors, T};
+
+mod func {
+	use hdx_parser::custom_function;
+	custom_function!(DynamicRangeLimitMix, atom!("dynamic-range-limit-mix"));
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize), serde())]
+pub struct DynamicRangeLimitMix<'a> {
+	function: T![Function],
+	values: Vec<'a, (T![Ident], T![Dimension::%], Option<T![,]>)>,
+	close: Option<T![')']>,
+}
+
+impl<'a> Peek<'a> for DynamicRangeLimitMix<'a> {
+	fn peek(p: &Parser<'a>) -> bool {
+		p.peek::<func::DynamicRangeLimitMix>()
+	}
+}
+
+impl<'a> Parse<'a> for DynamicRangeLimitMix<'a> {
+	fn parse(p: &mut Parser<'a>) -> ParserResult<Self> {
+		let function = p.parse::<T![Function]>()?;
+		let c = function.into();
+		if !func::DynamicRangeLimitMix::is(p, c) {
+			Err(diagnostics::Unexpected(c.into(), c.into()))?;
+		}
+		let mut values = Vec::new_in(p.bump());
+		loop {
+			if p.at_end() {
+				return Ok(Self { function, values, close: None });
+			}
+			if p.peek::<T![')']>() {
+				return Ok(Self { function, values, close: Some(p.parse::<T![')']>()?) });
+			}
+			let ident = p.parse::<T![Ident]>()?;
+			let length = p.parse::<T![Dimension::%]>()?;
+			let c: Cursor = length.into();
+			if !(0.0..=100.0).contains(&c.token().value()) {
+				Err(diagnostics::NumberOutOfBounds(c.token().value(), format!("{:?}", 0.0..=100.0), c.into()))?
+			}
+			let comma = p.parse_if_peek::<T![,]>()?;
+			values.push((ident, length, comma));
+		}
+	}
+}
+
+impl<'a> ToCursors<'a> for DynamicRangeLimitMix<'a> {
+	fn to_cursors(&self, s: &mut CursorStream<'a>) {
+		s.append(self.function.into());
+		for (ident, length, comma) in &self.values {
+			s.append(ident.into());
+			s.append(length.into());
+			if let Some(comma) = comma {
+				s.append(comma.into());
+			}
+		}
+		if let Some(close) = self.close {
+			s.append(close.into());
+		}
+	}
+}
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+	use crate::test_helpers::*;
+
+	#[test]
+	fn size_test() {
+		assert_size!(DynamicRangeLimitMix, 56);
+	}
+
+	#[test]
+	fn test_writes() {
+		assert_parse!(DynamicRangeLimitMix, "dynamic-range-limit-mix(high 80%,standard 20%)");
+		assert_parse!(DynamicRangeLimitMix, "dynamic-range-limit-mix(high 8%,standard 2%)");
+	}
+}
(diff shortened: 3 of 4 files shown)


crates/hdx_ast/src/css/values/color_hdr/impls.rs
```diff
@@ -1 +1,2 @@
 pub(crate) use crate::traits::StyleValue;
+pub(crate) use hdx_proc_macro::*;
```
keithamus, line 2: All `impls.rs` need to include at least these two lines so that the macros work.

crates/hdx_ast/src/css/values/color_hdr/mod.rs
```diff
@@ -8,12 +8,12 @@ use impls::*;
...
-// #[animation_type("by dynamic-range-limit-mix()")]
-// pub enum DynamicRangeLimit {}
+// https://drafts.csswg.org/css-color-hdr-1/#dynamic-range-limit
+#[value(" standard | high | constrained-high | <dynamic-range-limit-mix()> ")]
+#[initial("high")]
+#[applies_to("all elements")]
+#[inherited("no")]
+#[percentages("n/a")]
+#[canonical_order("per grammar")]
+#[animation_type("by dynamic-range-limit-mix()")]
```
keithamus, line 18: This is alll autogenerated. It doesn't need changing other than oncommenting.

crates/hdx_ast/src/css/values/color_hdr/mod.rs
```diff
@@ -8,12 +8,12 @@ use impls::*;
...
+pub enum DynamicRangeLimit<'a> {}
```
keithamus, line 19: A small change is needed here: the `<'a>` lifetime is needed because this enum depends on `DynamicRangeLimitMix` which itself has `<'a>` (because _it_ depends on a `Vec<'a>`).

crates/hdx_ast/src/css/values/color_hdr/types.rs
```diff
@@ -1,2 +1,78 @@
-pub(crate) use crate::css::types::*;
-pub(crate) use crate::css::units::*;
+use crate::css::units::LengthPercentage;
+use bumpalo::collections::Vec;
```
keithamus, line 1: We use `use bumpalo::collections::Vec` rather than `std::vec::Vec` so we can use bump allocations. They work the same other than bumpalo's vec requires a lifetime, so `Vec<'a, T>` rather than `Vec<T>`, in addition rather than `Vec::new()` we need `Vec::new_in(bump)`.

crates/hdx_ast/src/css/values/color_hdr/types.rs
```diff
@@ -1,2 +1,78 @@
...
+use hdx_parser::{diagnostics, CursorStream, Is, Parse, Parser, Peek, Result as ParserResult, ToCursors, T};
+
+mod func {
+	use hdx_parser::custom_function;
+	custom_function!(DynamicRangeLimitMix, atom!("dynamic-range-limit-mix"));
+}
```
keithamus, line 8: When comparing just one or two functions or idents, it can be more convenient to use `custom_function!` or `custom_keyword!` macros, which generate a mini parsable struct. They get a little unweildy if using many, so only use sparingly. I may end up removing this, depending on how much it gets used overall.

crates/hdx_ast/src/css/values/color_hdr/types.rs
```diff
@@ -1,2 +1,78 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
keithamus, line 10: All structs must derive all of these. Ideally `Copy` too but that's only possible when not depending on e.g. `Vec`.

crates/hdx_ast/src/css/values/color_hdr/types.rs
```diff
@@ -1,2 +1,78 @@
...
+#[cfg_attr(feature = "serde", derive(serde::Serialize), serde())]
```
keithamus, line 11: Also must depend on serde!

crates/hdx_ast/src/css/values/color_hdr/types.rs
```diff
@@ -1,2 +1,82 @@
...
+pub struct DynamicRangeLimitMix<'a> {
+	function: T![Function],
+	values: Vec<'a, (T![Ident], T![Dimension::%], Option<T![,]>)>,
+	close: Option<T![')']>,
```
keithamus, line 15: This struct is an AST node. It must comprise of all of the meaningful tokens made up of the grammar, such that it can re-constitute the original sources meaning by output the same tokens. In this case this struct has the following parts:

- `function` is the `T![Function]` token - this makes up the token of the function name up to and including the `(`. In the case of this token it'll always be the value `dynamic-range-limit-mix(`.
- `values` is a vec of Ident, % and optional trailing comma `,`. This is based on the grammar [defined in the spec](https://drafts.csswg.org/css-color-hdr-1/#dynamic-range-limit-mix) which is defined as `dynamic-range-limit-mix( [ <ident> && <percentage [0,100]> ]#)`. The `[ <<ident>> && <<percentage [0,100]>> ]` bit means an ident and a percentage _must occur, in order_. The `#` means that both pairs can occur one or more times, separated by commas. Hence `T![Ident] T![Dimension::%] Option<T![,]>`.
- `close` is an `Option<T![')']>`. This is the trailing `)` in the function. We make it an option because for our purposes we can be a little sloppy, and we don't _need_ the `)` to understand the node completely (I have plans later to do some form of error recovery and insert these tokens where possible, later on).

crates/hdx_ast/src/css/values/color_hdr/types.rs
```diff
@@ -1,2 +1,82 @@
...
+}
+
+impl<'a> Peek<'a> for DynamicRangeLimitMix<'a> {
+	fn peek(p: &Parser<'a>) -> bool {
+		p.peek::<func::DynamicRangeLimitMix>()
+	}
+}
```
keithamus, line 22: Every AST Node needs the `Peek` trait. The `Peek` trait should figure out how it can unambiguously determine if the next token is one that will start off this node, so we can cheaply check if we can parse this without having to cycle forward then rewind (which can be costly).

Here this is really easy, `p.peek::<func::DynamicRangeLimitMix>()` will say "the next token must be a function with the name `dynamic-range-limit-mix(`" which is pretty unambiguously going to be this node.

crates/hdx_ast/src/css/values/color_hdr/types.rs
```diff
@@ -1,2 +1,82 @@
...
+
+impl<'a> Parse<'a> for DynamicRangeLimitMix<'a> {
+	fn parse(p: &mut Parser<'a>) -> ParserResult<Self> {
+		let function = p.parse::<T![Function]>()?;
+		let c = function.into();
+		if !func::DynamicRangeLimitMix::is(p, c) {
+			Err(diagnostics::Unexpected(c.into(), c.into()))?;
+		}
```
keithamus, line 30: Here we check if the function is a Function token, but _isn't_ `dynamic-range-limit-mix`. If it's not we need to Err with some sensible error. For the most part `Unexpected` is fine, we can refine these at some later point.

crates/hdx_ast/src/css/values/color_hdr/types.rs
```diff
@@ -1,2 +1,82 @@
...
+		let mut values = Vec::new_in(p.bump());
+		loop {
+			if p.at_end() {
+				return Ok(Self { function, values, close: None });
+			}
```
keithamus, line 35: Loops should always check `p.at_end()` so they don't get stuck. Here you can also see we're returning a `None` for `close`.

crates/hdx_ast/src/css/values/color_hdr/types.rs
```diff
@@ -1,2 +1,82 @@
...
+			if p.peek::<T![')']>() {
+				return Ok(Self { function, values, close: Some(p.parse::<T![')']>()?) });
+			}
+			let ident = p.parse::<T![Ident]>()?;
+			let length = p.parse::<T![Dimension::%]>()?;
+			let c: Cursor = length.into();
+			if !(0.0..=100.0).contains(&c.token().value()) {
+				Err(diagnostics::NumberOutOfBounds(c.token().value(), format!("{:?}", 0.0..=100.0), c.into()))?
+			}
```
keithamus, line 44: Here we check to see if the % is within its bounds. We must check this because even though we can sometimes be sloppy, if the number is out of bounds we can't ever safely recover - because silently changing this to a value in bounds may go against the intent of the author (whereas, for example, adding a `)` token is probably quite safe).

crates/hdx_ast/src/css/values/color_hdr/types.rs
```diff
@@ -1,2 +1,82 @@
...
+	close: Option<T![')']>,
+}
+
+impl<'a> Peek<'a> for DynamicRangeLimitMix<'a> {
+	fn peek(p: &Parser<'a>) -> bool {
+		p.peek::<func::DynamicRangeLimitMix>()
+	}
+}
+
+impl<'a> Parse<'a> for DynamicRangeLimitMix<'a> {
```
keithamus, line 24: Every AST Node needs the `Parse` trait which is given a parser and can repeatedly parse other tokens or trees until it can complet its own node.

crates/hdx_ast/src/css/values/color_hdr/types.rs
```diff
@@ -1,2 +1,82 @@
...
+			if !(0.0..=100.0).contains(&c.token().value()) {
+				Err(diagnostics::NumberOutOfBounds(c.token().value(), format!("{:?}", 0.0..=100.0), c.into()))?
+			}
+			let comma = p.parse_if_peek::<T![,]>()?;
+			values.push((ident, length, comma));
+		}
+	}
+}
+
+impl<'a> ToCursors<'a> for DynamicRangeLimitMix<'a> {
```
keithamus, line 51: Every AST node must also implement the `ToCursors` trait which is given a `CursorStream` and must feed its cursors, in order, to the cursor stream. This then lets us write the cursor stream back out to a string.

crates/hdx_ast/src/css/values/color_hdr/types.rs
```diff
@@ -1,2 +1,82 @@
...
+	fn to_cursors(&self, s: &mut CursorStream<'a>) {
+		s.append(self.function.into());
```
keithamus, line 53: Something like a `T![Function]` is a "newtyped" Cursor, so calling `.into()` should convert it into a Cursor. For things made up of multiple cursors you'll need to call `ToCursors::to_cursors(&self.the_thing, s)`.

crates/hdx_ast/src/css/values/color_hdr/types.rs
```diff
@@ -1,2 +1,82 @@
...
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+	use crate::test_helpers::*;
+
+	#[test]
+	fn size_test() {
+		assert_size!(DynamicRangeLimitMix, 56);
+	}
```
keithamus, line 75: Important to track sizes of all structs so we can figure out places to compress or align or otherwise optimize hot nodes.

crates/hdx_proc_macro/src/def.rs
```diff
@@ -1350,8 +1350,8 @@ impl DefType {
 
 	pub fn requires_allocator_lifetime(&self) -> bool {
 		if let Self::Custom(DefIdent(ident), _) = self {
-			return matches!(ident, &atom!("OutlineColor") | &atom!("BorderTopColor") | &atom!("AnchorName"));
-		}
+			return matches!(ident, &atom!("OutlineColor") | &atom!("BorderTopColor") | &atom!("AnchorName") | &atom!("DynamicRangeLimitMix"));
```
keithamus, line 1353: Because `DynamicRangeLimitMix` required the `<'a>` lifetime, it needs to be added here. This is... not ideal, but it works for now. The generator needs to be smarter.

keithamus (commented):
This might serve as a good example of how to implement some values, and so I'll leave some commentary regarding this PR:

PR #62, opened 2024-12-12
Start implementing `<font-weight-absolute>`

koddsson:
Parking this here because I don't know what I'm doing 🫠


crates/hdx_ast/src/css/values/fonts/types.rs
```diff
@@ -1 +1,79 @@
...
+impl<'a> Parse<'a> for FontWeightAbsolute {
+	fn parse(p: &mut Parser<'a>) -> ParserResult<Self> {
+		let ident = p.parse::<T![Ident]>()?;
+		let c: Cursor = ident.into();
+		match p.parse_atom_lower(c) {
+			atom!("normal") => Ok(Self::Normal(<T![Ident]>::build(p, c))),
+			atom!("bold") => Ok(Self::Bold(<T![Ident]>::build(p, c))),
+			atom!("bolder") => Ok(Self::Bolder(<T![Ident]>::build(p, c))),
+			atom!("lighter") => Ok(Self::Lighter(<T![Ident]>::build(p, c))),
+			atom => Err(diagnostics::UnexpectedIdent(atom, c.into()))?,
```
koddsson: I feel like this should have a `Number => Ok(...)` arm here as well but wasn't sure how to do that.

crates/hdx_ast/src/css/values/fonts/types.rs
```diff
@@ -1 +1,79 @@
...
+	#[test]
+	fn test_writes() {
+		assert_parse!(FontWeightAbsolute, "normal");
+		assert_parse!(FontWeightAbsolute, "bold");
+		assert_parse!(FontWeightAbsolute, "bolder");
+		assert_parse!(FontWeightAbsolute, "lighter");
+		assert_parse!(FontWeightAbsolute, "100");
+		assert_parse!(FontWeightAbsolute, "500");
+		assert_parse!(FontWeightAbsolute, "900");
+	}
```
koddsson: Can I run the tests for just this file in isolation so I can quickly get just warnings and errors that relate to this change? I'm a bit overwhelmed with all the output when I run `cargo test` :D

crates/hdx_ast/src/css/values/fonts/types.rs
keithamus: You’ll need something like

```rust
if p.peek::<T![Ident]>() { 
    …
} else if p.peek::<T![Number]>() {
    …
}
```

crates/hdx_ast/src/css/values/fonts/types.rs
keithamus: `cargo test -p hdx_ast` scopes to just the hdx_ast` package, but you can also filter by module or test name so in the case running `cargo test -p hdx_ast css::values::fonts` should scope it to just these tests.

crates/hdx_ast/src/css/values/fonts/types.rs
```diff
@@ -1 +1,79 @@
-pub(crate) use crate::css::units::*;
+use hdx_atom::atom;
+use hdx_lexer::Cursor;
+use hdx_parser::{diagnostics, Build, CursorStream, Parse, Parser, Peek, Result as ParserResult, ToCursors, T};
+
+mod func {
```
keithamus: This is, by convention, `mod kw`.

But I'd probably get rid of all the keywords and just peek an ident.

keithamus (commented):
Looking great so far!

keithamus (approved):

keithamus:
Thanks @koddsson! I updated this and merged it.

Issue #66, opened 2024-12-15
Investigate why Windows tests fail


keithamus:
Looks like windows tests are failing: https://github.com/keithamus/hdx/actions/runs/12334550029/job/34424890161

Seems like offsets are a few characters off in the snapshot tests. Is this CRLF line ending? Can we avoid this somehow?

keithamus:
Windows still failing.

PR #73, opened 2024-12-17
chore(ci): automatically upload binaries to github release's

maraisr:
What this pull request does is publish the latest binaries to the `canary` git tag release. GitHub can house release assets per tag, or per release issue.

Until we hit something stable, we should publish the latest binaries to the `canary` tag. 

There is an additional change, in that the release action will only run if the `test` workflow succeeded. I also got rid of the "release tag" or issue creation flows, probably we can explore that closer to when we wanna start tagging properly. I've found that using [git tags to "trigger" releases](https://github.com/maraisr/dldr/blob/614317540d4982b779a3756a560a96b74ec2060f/.github/workflows/ci.yml#L31), work really well. But given we embed the css spec in the binary, id probably like to see us split the "db" from the cli, such that we do not need a new NPM/extension release just to capture new specs. And instead do something like caniuse, in that the db is a separate binary. But lets discuss this some other time, there are pros in a self contained NPM too.

For now, this will publish the binaries to a canary release, we can then use to power things like a Zed extension.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,10 +1,10 @@
 # mostly copied from https://raw.githubusercontent.com/web-infra-dev/oxc/main/.github/workflows/release_cli.yml
 name: release
 on:
-  push:
-    branches: ["main"]
-  release:
-    types: [created]
+  workflow_run:
+    workflows: [test]
+    types: [completed]
+    branches: [main]
   workflow_dispatch:
 concurrency:
   group: "release"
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -1,10 +1,10 @@
 # mostly copied from https://raw.githubusercontent.com/web-infra-dev/oxc/main/.github/workflows/release_cli.yml
 name: release
 on:
-  push:
-    branches: ["main"]
-  release:
-    types: [created]
+  workflow_run:
+    workflows: [test]
```
keithamus, line 5: This will require us to fix #66 before the next release, right?

.github/workflows/release.yml
```diff
@@ -13,7 +13,29 @@ permissions:
...
+    runs-on: ubuntu-latest
+    needs: [build]
+    steps:
+      - uses: actions/checkout@v4
+      - name: Tag
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git tag -fa canary -m "Latest Continuous Release" ${GITHUB_SHA}
+          git push --force origin canary
```
keithamus: Is this worth a `--force-with-lease` just in case? Maybe also just push the canary tag and nothing else?

keithamus (approved):
Looks awesome. Couple of nits:

.github/workflows/release.yml
maraisr: Yeah, think so. I tried to solve this on my local windows machine, and got it to test and build with this diff: https://github.com/keithamus/hdx/pull/74 which I can see is failing now with some CRLF things maybe which is strange, because did not locally.

PR #74, opened 2024-12-17
fix: we cannot pprof in windows

### LINKED ISSUE #66: Investigate why Windows tests fail

maraisr:
`pprof` is a "unix" only dependency. Even though we do not "use" it anywhere not protected by `cfg!`, it seems as though `cargo test` builds all dependencies, which `pprof` fails. 

Fix is simple, simply only include that dep on `unix` systems. Confirmed testing/building on my Windows machine.

closes: https://github.com/keithamus/hdx/issues/66

### DIFF
--- a/crates/hdx_lsp/Cargo.toml
+++ b/crates/hdx_lsp/Cargo.toml
@@ -38,10 +38,12 @@ tracing-subscriber = { workspace = true }
 [dev-dependencies]
 glob = { workspace = true }
 criterion = { workspace = true, features = ["html_reports"] }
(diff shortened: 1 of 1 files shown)


keithamus (approved):
I don't think I have setup benchmarks in the lsp package yet, so this is copypasta, but nevertheless we'll probably bench soon so let's merge this!

PR #75, opened 2024-12-17
feat: mvp zed extension going

maraisr:
Get's a Zed Extension impl going. Where the LSP is downloaded from GitHub release assets.

Even though Zed extensions are Rust projects, we shouldn't include that as part of our `hdx` workspace. Furthermore, we should treat the lsp as external, and not try to compile our lsp directly within the extension wasm.

The `lsp` location is resolved through either;

1. env: `HDX_SERVER_PATH` (which is the same as the VSCode one)
1. we download it from GitHub's release assets

CleanShot 2024-12-17 at 18 00 26

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,7 +23,7 @@ jobs:
           git config user.name "github-actions[bot]"
           git config user.email "github-actions[bot]@users.noreply.github.com"
           git tag -fa canary -m "Latest Continuous Release" ${GITHUB_SHA}
-          git push --force-with-lease origin canaray:refs/tags/canary
+          git push --force-with-lease origin canary:refs/tags/canary
 
   build:
     if: |
@@ -96,13 +96,14 @@ jobs:
           BIN_NAME=hdx-${{ matrix.code-target }}
           mv target/${{ matrix.target }}/release/hdx $BIN_NAME
 
+      - run: chmod +x hdx-*
+
       - name: Upload Binary
         uses: actions/upload-artifact@v4
         with:
           if-no-files-found: error
           name: binary-${{ matrix.code-target }}
-          path: |
-            hdx-*
+          path: hdx-*
 
   deploy:
     env:
@@ -121,7 +122,7 @@ jobs:
           cache: "npm"
       - name: Download Artifacts
         uses: actions/download-artifact@v4
-      - run: cp binary-*/* packages/hdx/bin && chmod +x packages/hdx/bin/*
+      - run: cp binary-*/* packages/hdx/bin
       - run: tree
       - run: npm i
         working-directory: ./packages/hdx
--- a/packages/hdx_vscode/package.json
+++ b/packages/hdx_vscode/package.json
@@ -9,7 +9,7 @@
 		"url": "https://github.com/sponsors/keithamus"
 	},
 	"license": "MIT",
-	"author": "Keith Cirkel (https://keithcirke.co.uk)",
+	"author": "Keith Cirkel (https://keithcirkel.co.uk)",
 	"type": "commonjs",
 	"main": "./src/main.js",
 	"files": [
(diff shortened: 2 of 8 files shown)


packages/hdx_zed/dev.md
```diff
@@ -0,0 +1,12 @@
+To install the dev version of the zed extension:
+
+1. `rustup target add wasm32-wasip1`
+2. Extensions > Install Dev Extgension > Select the `pacakges/hdx_zed` folder
```
keithamus:
```suggestion
2. Extensions > Install Dev Extension > Select the `pacakges/hdx_zed` folder
```

packages/hdx_zed/extension.toml
```diff
@@ -0,0 +1,11 @@
+id = "hdx"
+name = "hdx"
+version = "0.0.1"
+schema_version = 1
+authors = ["Keith Cirkel (https://keithcirke.co.uk)"]
```
keithamus: You should add your name here! But also
```suggestion
authors = ["Keith Cirkel (https://keithcirkel.co.uk)"]
```

packages/hdx_zed/extension.toml
```diff
@@ -0,0 +1,11 @@
...
+description = "A refreshing new css experience"
```
keithamus:
```suggestion
description = "Refreshing CSS!"
```

packages/hdx_zed/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+			.find(|asset| asset.name == asset_name)
+			.ok_or_else(|| format!("no asset found matching {:?}", asset_name))?;
+
+		if !fs::metadata(&asset_name).map_or(false, |stat| stat.is_file()) {
+			zed::set_language_server_installation_status(
+				language_server_id,
+				&zed::LanguageServerInstallationStatus::Downloading,
+			);
+
+			zed::download_file(&asset.download_url, "bin", zed::DownloadedFileType::Zip)
```
maraisr: Once we have the release assets going, this line may need to change a tad. As in is it `bin/binary-platform-arch`??

packages/hdx_zed/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+	fn language_server_binary_path(
+		&mut self,
+		language_server_id: &zed_extension_api::LanguageServerId,
+		worktree: &zed::Worktree,
+	) -> Result<String> {
+		if let Some(path) = worktree.which("hdx") {
+			return Ok(path);
+		}
+
+		if let Some(path) = &self.cached_binary_path.clone().or_else(|| env::var("HDX_SERVER_PATH").ok()) {
```
keithamus: This line might move up so the env var always overrides `which`.

keithamus (approved):
Looks amazing! Couple of small things:

packages/hdx_zed/extension.toml
maraisr: I actually copied this from the vscode package json. So possibly typo there too, will fix up. 

And yeah dude, thank you will do! 🥰

packages/hdx_vscode/package.json
```diff
@@ -9,7 +9,7 @@
 		"url": "https://github.com/sponsors/keithamus"
 	},
 	"license": "MIT",
-	"author": "Keith Cirkel (https://keithcirke.co.uk)",
```
keithamus, line 12: :pray:

.github/workflows/release.yml
```diff
@@ -96,13 +96,14 @@ jobs:
           BIN_NAME=hdx-${{ matrix.code-target }}
           mv target/${{ matrix.target }}/release/hdx $BIN_NAME
 
+      - run: chmod +x hdx-*
```
maraisr, line 99: I needed `zed::make_file_executable` in the extension, which we don't need to do; if we `chmod +x` the binary. 

I believe these file modes are transfered through the assets; but wasnt too sure so kept the zed fn api to `+x` it.

PR #108, opened 2025-06-26
Fix windows ci builds

### LINKED ISSUE #66: Investigate why Windows tests fail

maraisr:
Windows builds always used to work fine for me locally, because my .gitconfig has `autocrlf = input`. But it seems as though there are some line-endings in our fixtures that may not always be `lf` like they are when we generate snapshots, presumably on Linux/Mac.

So this PR simply tells CI, hey autocrlf first. And yay, things pass!! I also added a .gitattributes to force `lf` line-endings on all source files.

Here is a sample of CI passing on Windows: https://github.com/maraisr/hdx/actions/runs/15894577728/job/44823352507

closes: https://github.com/csskit/csskit/issues/66

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,6 +28,9 @@ jobs:
     name: Build & Test on ${{ matrix.code-target }}
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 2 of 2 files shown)


keithamus (approved):
Excellent!

PR #113, opened 2025-06-30
Implements ordered combinators and uncomments some align properties

maraisr:
Implements Ordered Combinators def values and uncomments some align properties as it can now be parsed with this addition.

It currently only supports optional types, ie `<color>?` and not optional idents, like `auto?`. There are a handful of properties that make use of optional idents in combinators, but want to handle that separately.

There are probably more properties we can uncomment with this change, but feel that can possibly be handled in a followup.

### DIFF
--- a/crates/css_ast/src/values/align/mod.rs
+++ b/crates/css_ast/src/values/align/mod.rs
@@ -6,15 +6,16 @@ use impls::*;
  * CSS Box Alignment Module Level 3
  */
 
-// // https://drafts.csswg.org/css-align-3/#align-content
-// #[value(" normal | <baseline-position> |  <content-distribution> | <overflow-position>? <content-position> ")]
-// #[initial("normal")]
-// #[applies_to("block containers, multicol containers, flex containers, and grid containers")]
-// #[inherited("no")]
-// #[percentages("n/a")]
-// #[canonical_order("per grammar")]
-// #[animation_type("discrete")]
-// pub enum AlignContentStyleValue {}
+// https://drafts.csswg.org/css-align-3/#align-content
+// #[value(" normal | <baseline-position> | <content-distribution> | <overflow-position>? <content-position> ")]
+#[value(" normal | <baseline-position> | <content-distribution> | <content-position> ")]
+#[initial("normal")]
+#[applies_to("block containers, multicol containers, flex containers, and grid containers")]
+#[inherited("no")]
+#[percentages("n/a")]
+#[canonical_order("per grammar")]
+#[animation_type("discrete")]
+pub enum AlignContentStyleValue {}
 
 // // https://drafts.csswg.org/css-align-3/#justify-content
 // #[value(" normal | <content-distribution> | <overflow-position>? [ <content-position> | left | right ] ")]
--- a/crates/css_ast/src/values/align/mod.rs
+++ b/crates/css_ast/src/values/align/mod.rs
@@ -7,8 +7,7 @@ use impls::*;
  */
 
 // https://drafts.csswg.org/css-align-3/#align-content
-// #[value(" normal | <baseline-position> | <content-distribution> | <overflow-position>? <content-position> ")]
-#[value(" normal | <baseline-position> | <content-distribution> | <content-position> ")]
+#[value(" normal | <baseline-position> | <content-distribution> | <overflow-position>? <content-position> ")]
 #[initial("normal")]
 #[applies_to("block containers, multicol containers, flex containers, and grid containers")]
 #[inherited("no")]
@@ -27,15 +26,15 @@ pub enum AlignContentStyleValue {}
 // #[animation_type("discrete")]
 // pub enum JustifyContentStyleValue {}
 
-// // https://drafts.csswg.org/css-align-3/#place-content
-// #[value(" <'align-content'> <'justify-content'>? ")]
-// #[initial("normal")]
-// #[applies_to("block containers, flex containers, and grid containers")]
-// #[inherited("no")]
-// #[percentages("n/a")]
-// #[canonical_order("per grammar")]
-// #[animation_type("discrete")]
-// pub struct PlaceContentStyleValue;
+// https://drafts.csswg.org/css-align-3/#place-content
+#[value(" <'align-content'> <'justify-content'>? ")]
+#[initial("normal")]
+#[applies_to("block containers, flex containers, and grid containers")]
+#[inherited("no")]
+#[percentages("n/a")]
+#[canonical_order("per grammar")]
+#[animation_type("discrete")]
+pub struct PlaceContentStyleValue;
 
 // // https://drafts.csswg.org/css-align-3/#justify-self
 // #[value(" auto | normal | stretch | <baseline-position> | <overflow-position>? [ <self-position> | left | right ] ")]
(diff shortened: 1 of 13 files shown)


crates/css_ast/src/values/flexbox/mod.rs
```diff
@@ -114,4 +114,4 @@ pub enum AlignSelfStyleValue {}
 #[percentages("n/a")]
 #[canonical_order("per grammar")]
 #[animation_type("discrete")]
-pub enum AlignContentStyleValue {}
+pub enum FlexAlignContentStyleValue {}
```
maraisr: @keithamus I wasnt too sure here. The names had a conflict with what css_ast::values::align was trying to expose. Not sure which is the canonical truth, or how we handle conflicts like this. Just prepended to keep moving, but fully open to your advice.

I almost feel as though we should delete this one

crates/css_ast/src/values/align/mod.rs
```diff
@@ -6,15 +6,16 @@ use impls::*;
...
-// #[initial("normal")]
-// #[applies_to("block containers, multicol containers, flex containers, and grid containers")]
-// #[inherited("no")]
-// #[percentages("n/a")]
-// #[canonical_order("per grammar")]
-// #[animation_type("discrete")]
-// pub enum AlignContentStyleValue {}
+// https://drafts.csswg.org/css-align-3/#align-content
+// #[value(" normal | <baseline-position> | <content-distribution> | <overflow-position>? <content-position> ")]
+#[value(" normal | <baseline-position> | <content-distribution> | <content-position> ")]
```
maraisr: Yeah, wasnt too sure how to handle `<overflow-position>?`. So omitted for now. Can you give me a hint here @keithamus

crates/csskit_proc_macro/src/def.rs
```diff
@@ -388,6 +389,11 @@ impl Def {
 				let ident = format_ident!("Literal{}", variant_str);
 				quote! { #ident }
 			},
+			Self::Combinator(ds, DefCombinatorStyle::Ordered) => {
```
maraisr, line 392: Just pick the first non optional def, and fallback to the first optional. Not sure if this will be elegant to use for 100% of the cases, but does look kinda nice now:

```
"<color>? auto"
Auto((Optional<Color>, Ident))
```

crates/csskit_proc_macro/src/def.rs
```diff
@@ -1271,7 +1299,27 @@ impl GeneratePeekImpl for Def {
 			Self::Ident(p) => p.peek_steps(),
 			Self::Function(_, _) => quote! { <::css_parse::T![Function]>::peek(p, c) },
 			Self::Optional(p) => p.peek_steps(),
-			Self::Combinator(p, DefCombinatorStyle::Ordered) => p[0].peek_steps(),
+			Self::Combinator(ds, DefCombinatorStyle::Ordered) => {
```
maraisr, line 1302: For `AlignContentStyleValue`, if its value was just `flex-end` (we have a test for that now) — simply peeking `[0]` would fail, because the first def is `<overflow-position>? => safe | unsafe`, we'd never parse the `content-position`. 

So, peek all optional values, and one required def.

There is probably a way to reduce the code duplication here

crates/csskit_proc_macro/src/def.rs
```diff
@@ -1447,6 +1495,23 @@ impl GenerateParseImpl for Def {
 					}
 				}
 			}
+			Self::Combinator(ds, DefCombinatorStyle::Ordered) => {
+				let inner = capture.unwrap_or_else(|| format_ident!("val"));
+				let idents: Vec<Ident> = (0..ds.len()).map(|i| format_ident!("{}{}", inner, i)).collect();
+
+				let steps: Vec<TokenStream> = ds.iter().enumerate()
```
maraisr, line 1514: As youll see in the tests, the algorithm here is:

- if its an optional value, `parse_if_peek`
- otherwise `parse`

and collect into variables:

```
<color>? keyword

let val0 = parse_if_peek(color)
let val1 = parse(keyword)
let val = (val0, val1)

Ok(Enum::Item(val))
```

crates/csskit_proc_macro/src/def.rs
```diff
@@ -1168,16 +1197,15 @@ impl GenerateToCursorsImpl for Def {
...
 			}
-			Self::Combinator(opts, DefCombinatorStyle::Ordered) => {
-				let exprs: Vec<TokenStream> = (0..opts.len())
-					.map(|i| {
+			Self::Combinator(ds, DefCombinatorStyle::Ordered) => {
+				let exprs: Vec<TokenStream> = ds.iter().enumerate()
+					.map(|(i, def)| {
 						let index = Index { index: i as u32, span: Span::call_site() };
-						quote! { ::css_parse::ToCursors::to_cursors(&self::#index, s); }
+						def.to_cursors_steps(quote!{ &#capture.#index })
```
maraisr, line 1204: Hmmm, wasnt too sure about this @keithamus. When we do `to_cursor_steps`, in roughly line 1241, for alternative combinators, it calls into here.

But the signature of alternative combinators, is different to im assuming straight up ordered combinator.

`enum` vs `struct`. 

I feel this change is more than fine, given `#capture` ought to be `self` for the struct syntax, and `inner` for alternative combinators. Ie, just use the capture.

Tests pass, so think we're good?

crates/csskit_proc_macro/src/def.rs
```diff
@@ -1705,7 +1770,7 @@ impl GenerateParseImpl for DefIdent {
 		quote! {
 			let #capture = p.parse::<::css_parse::T![Ident]>()?;
 			let c: ::css_lexer::Cursor = #capture.into();
-			p.parse_str_lower(t);
+			// p.parse_str_lower(t);
```
maraisr: 🤷 No idea if this line is required. But compiling a test like: `foo | <color>? bar`, gets into this `DefIdent::parse_steps` method, and had no idea what `t` was, or how to resolve it, in the combinator. 

I know it wants a cursor, so wasnt sure if it can just be `p.parse_str_lower(c);` (from one line above).

So just commented out for now, tests pass. But keen to hear what i should do instead @keithamus

crates/csskit_proc_macro/src/snapshots/csskit_proc_macro__test__group_multiple_optional.snap
```diff
@@ -0,0 +1,67 @@
...
+#[automatically_derived]
+impl ::css_parse::ToCursors for Foo {
+    fn to_cursors(&self, s: &mut impl ::css_parse::CursorSink) {
+        match self {
+            Self::Foo(inner) => {
+                s.append(inner.into());
+            }
+            Self::Bar(inner) => {
+                if let Some(inner) = &inner.0 {
+                    ::css_parse::ToCursors::to_cursors(inner, s);
```
maraisr, line 57: I didnt touch this code, but explored it. `to_cursors` just does the `s.append(inner.into())`. Wasnt sure if there is an appetite, to replace all the `s.append(inner.into())` generated code to `::css_parse::ToCursors::to_cursors(inner, s)`, so there is a canonical place for how "to cursors" stuff is done. 

Follow up pr maybe.

crates/csskit_proc_macro/src/def.rs
keithamus: Let's just remove it, it's probably a holdover. 

`parse_str_lower` should probably have `#[must_use]` - there are probably a lot of methods that could use this.

PR #115, opened 2025-07-01
Implements `glyph-orientation-vertical` along with IntLiteral and DimensionLiteral value defs

maraisr:
Implements `IntLiteral` and `DimensionLiteral` def values and uncomments `glyph-orientation-vertical` as it can now be parsed with this addition.

There were some changes in `DimensionKind` to support converting a `str` to its enum value, and some clean ups in how keywords emit their `parse_steps`.


keithamus (approved):
Nice!

PR #135, opened 2025-07-05
Implements some font style values

maraisr:
Uncomments `FontStyleStyleValue` and `FontSizeStyleValue`. Also got a start on `FontPaletteStyleValue` with `<palette-identifier>`.

There was a slight change in csskit_proc_macros with respects to borrowing that unlocked this.


crates/css_ast/src/values/fonts/impls.rs
```diff
@@ -1,2 +1,48 @@
...
+		// assert_eq!(std::mem::size_of::<FontVariationSettingsStyleValue>(), 1);
+		// assert_eq!(std::mem::size_of::<FontPaletteStyleValue>(), 1);
+		assert_eq!(std::mem::size_of::<FontVariantEmojiStyleValue>(), 16);
+	}
+
+	#[test]
+	fn test_writes() {
+		assert_parse!(FontStyleStyleValue, "normal");
+		assert_parse!(FontStyleStyleValue, "oblique 45deg");
+		assert_parse_error!(FontStyleStyleValue, "oblique 45px");
```
maraisr, line 43: I originally had a test in for `95deg`, which would sit outside of the `<angle [-90deg,90deg]>` — but that didn't cause a parse error. I think I understand why it wouldn't, but question—do we retain the ranges anywhere in the type? I can see this be useful for things like minifies which could clamp values on output.

crates/csskit_proc_macro/src/snapshots/csskit_proc_macro__test__group_optional_last.snap
```diff
@@ -0,0 +1,61 @@
...
+}
+#[automatically_derived]
+impl ::css_parse::ToCursors for Foo {
+    fn to_cursors(&self, s: &mut impl ::css_parse::CursorSink) {
+        match self {
+            Self::Foo(inner) => {
+                ::css_parse::ToCursors::to_cursors(inner, s);
+            }
+            Self::Bar(inner) => {
+                ::css_parse::ToCursors::to_cursors(&inner.0, s);
```
maraisr, line 54: Prior to changing this to `ToCursors::to_cursors`, this compilation would fail — because `s.append(&inner.0.into())` would I think consume, the borrow? Idk. Doing this fixed it

crates/csskit_proc_macro/src/def.rs
```diff
@@ -1811,7 +1811,7 @@ impl DefIdent {
 
 impl GenerateToCursorsImpl for DefIdent {
 	fn to_cursors_steps(&self, inner: TokenStream) -> TokenStream {
-		quote! { s.append(#inner.into()); }
+		quote! { ::css_parse::ToCursors::to_cursors(#inner, s); }
```
maraisr, line 1814: This unlocked `FontStyleStyleValue` 🎉

crates/css_ast/tests/snapshots/popular_snapshots__popular_mini.snap
```diff
@@ -1190,19 +1181,15 @@ snapshot_kind: text
...
             },
             "value": {
-              "values": [
-                {
-                  "kind": "Whitespace",
-                  "offset": 977,
-                  "len": 1
-                },
-                {
+              "LengthPercentage": {
```
maraisr, line 1184: This felt interesting, 🤔

crates/css_ast/src/types/palette_identifier.rs
```diff
@@ -0,0 +1,4 @@
+use css_parse::T;
+
+// https://www.w3.org/TR/css-fonts-4/#typedef-font-palette-palette-identifier
+pub type PaletteIdentifier = T![DashedIdent];
```
keithamus, line 4: It would be helpful to use the [new type idiom](https://doc.rust-lang.org/rust-by-example/generics/new_types.html) here, because then we can `impl` on top of this (for example, later on we can track whether or not a PaletteIdentifier is used in an `@font-face` rule). I believe the current `pub type` will fall away meaning we cannot distinguish between `T![DashedIdent]` and a `PaletteIdentifier`.


```suggestion
pub struct PaletteIdentifier(T![DashedIdent]);
```

This will mean implementing `parse` etc but perhaps we can add a `derive(Parse)` for such simple wrappers?

crates/css_ast/src/values/fonts/impls.rs
keithamus: > didn't cause a parse error

Yes this should be a parse error. IMO anything where outputting it 1-1 would cause a real CSS engine to discard the rule should be a parse error.

> do we retain the ranges anywhere in the type?

The validation? No that's handled in `def.rs` (or a hand rolled parse impl).

The written value is stored in the token, so something like `self.0.token().value()` should get you the number of the dimension.

keithamus (approved):
This LGTM as is, but I think there is some room for improvement.

crates/css_ast/src/values/fonts/impls.rs
maraisr: Yeh, okay makes sense. I'll see if we can generate specific parse_steps when a type has a def range. Will try

crates/css_ast/src/types/palette_identifier.rs
maraisr: I might undo this from this PR, and tackle that in a separate diff

PR #136, opened 2025-07-05
Correctly handle parsing optional bounded types

maraisr:
Follows on from https://github.com/csskit/csskit/pull/135, to now correctly parse invalid optional bounded types, and adds tests for it.

Also renamed some proc_macro tests, named better. `a | b` is not a "group" for example.


crates/csskit_proc_macro/src/snapshots/csskit_proc_macro__test__bounded_range_multiplier_is_optimized_to_options.snap
```diff
@@ -22,8 +21,18 @@ impl<'a> ::css_parse::Parse<'a> for Foo {
     fn parse(p: &mut ::css_parse::Parser<'a>) -> ::css_parse::Result<Self> {
         use ::css_parse::{Parse, Peek};
         let val0 = p.parse::<crate::AnimateableFeature>()?;
-        let val1 = p.parse_if_peek::<crate::AnimateableFeature>()?;
-        let val2 = p.parse_if_peek::<crate::AnimateableFeature>()?;
+        let val1 = if p.peek::<crate::AnimateableFeature>() {
```
maraisr, line 24: Yep, since we rely on a _generic_ `parse_steps` method, this code looks a little funky, given `parse_if_peek`. But I think functionally they are equivalent. And should a specific parse_steps do more (say bounded Angles), we'll correctly handle those.

crates/css_ast/src/values/fonts/impls.rs
```diff
@@ -41,6 +41,7 @@ mod tests {
 		assert_parse!(FontStyleStyleValue, "normal");
 		assert_parse!(FontStyleStyleValue, "oblique 45deg");
 		assert_parse_error!(FontStyleStyleValue, "oblique 45px");
+		assert_parse_error!(FontStyleStyleValue, "oblique 91deg");
```
keithamus: Nit: Something I overlooked in the previous PR, can we add a `fn test_errors()` just for the error conditions, that way it's easier to read the two. Maybe also a comment above each parse error explaining why.

keithamus (approved):

keithamus (approved):

PR #138, opened 2025-07-05
Reintroduce derive(Parse/Peek/ToCursors)

keithamus:
We used to have `derive(Parse)` & `derive(Peek)` but these were deemed unnecessary when we introduced `csskit_proc_macro`. Since then we've added a lot of new types in `src/types/` and some of these have been simple enough (including some new type structs) that they could benefit from deriving Parse, Peek or even ToCursors.

This PR re-introduces `Parse` & `Peek` derive macros, which simply call `parse` on their underlying types. This also introduces the `ToCursors` derive macro. 

All of these now mean for trivial impls of these, we can simply `#[derive(Peek, Parse, ToCursors)]`.

Of course for single token impls `Build` & `Into<Cursor>` may be better.

This also removes a bunch of trivial implementations in favour of using the derives, which also helped prove out these implementations!

### DIFF
--- a/crates/css_ast/src/types/bg_image.rs
+++ b/crates/css_ast/src/types/bg_image.rs
@@ -1,11 +1,12 @@
 use css_lexer::Cursor;
-use css_parse::{CursorSink, Parse, Parser, Peek, Result as ParserResult, T, ToCursors, diagnostics};
+use css_parse::{Parse, Parser, Peek, Result as ParserResult, T, diagnostics};
+use csskit_derives::ToCursors;
 
 use super::Image;
 
 // https://drafts.csswg.org/css-backgrounds/#typedef-bg-image
 // <bg-image> = <image> | none
-#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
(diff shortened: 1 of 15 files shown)


crates/css_ast/src/types/bg_image.rs
```diff
@@ -34,11 +35,19 @@ impl<'a> Parse<'a> for BgImage<'a> {
 	}
 }
 
-impl ToCursors for BgImage<'_> {
```
maraisr, line 37: oh yeah so much nicer

crates/csskit_derives/src/to_cursors.rs
```diff
@@ -0,0 +1,73 @@
...
+				})
+				.collect();
+			quote! { #(#steps)* }
+		}
+
+		Data::Struct(_) => err(ident.span(), "Cannot derive ToCursors on a struct with named fields"),
+
+		Data::Union(_) => err(ident.span(), "Cannot derive ToCursors on a Union"),
+
+		Data::Enum(DataEnum { variants, .. }) => {
```
maraisr, line 31: Would be cool to have one `assert_snapshot!` test, makes seeing what this compiled code looks like, much easier.

maraisr (approved):
Yeah this is nice, may speed up some of the more trivial types yeah

PR #162, opened 2025-07-07
(css_ast): Implements `transform`

### LINKED ISSUE #140: Todo: <transform-list>
[`<transform-list>`](https://github.com/csskit/csskit/blob/main/crates/css_ast/src/types/transform_list.rs) is a stub, but it's needed for the `transform` property. Its an array of TransformFunctions, which themselves are an enum of a whole bunch of transform methods, e.g. `rotate()`, `translate()` and so on. Each enum has its own syntax but they could each be iterated on independantly, and each one would be marked as `todo!()`, to reduce scope on what's needed.

https://drafts.csswg.org/css-transforms-1/#typedef-transform-list
https://drafts.csswg.org/css-transforms-1/#two-d-transform-functions
```
<transform-list> = <transform-function>+
```

[80% of web pages](https://chromestatus.com/metrics/css/timeline/popularity/446) use this rule.

maraisr:
closes: https://github.com/csskit/csskit/issues/140


crates/css_ast/src/types/transform/matrix.rs
```diff
@@ -0,0 +1,111 @@
+use css_lexer::Cursor;
+use css_parse::{CursorSink, Parse, Parser, Peek, Result as ParserResult, T, ToCursors, diagnostics};
+
+// https://drafts.csswg.org/css-transforms-1/#funcdef-transform-matrix
+// matrix() = matrix( <number>#{6} )
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize), serde())]
+pub struct Matrix {
```
maraisr: I wanna chat to you @keithamus about some kind of new `function_def!` macro.

Matrix, and all other fns for transform really, can just utilize `derive(Parse/Peek/ToCursors)`. They are simple enough. I mean, not really Parse or Peek—given we need to look for specific keywords.

Not unsure how working with the ast will be, but to me at least `matrix.0` isnt meaningful. Going `matrix.a` is much more useful. I kinda wanna create a macro that looks like:


```rs
function_def!(pub struct Matrix {
  pub a: T![Number],
  ...
})
```

and have that expand into something like:

```rs
struct Matrix {
  pub a: T![Number],
  ...
  _a_comma: Option<T![',']>,
  _fn: T![Function],
  _close: Option<T![')']>,
```

with peek and parse steps defined for you, like checking that the function name is `lower(struct_name)` etc.

If youre open to this idea, im happy to propose it.

Originally I thought, "oh hey fns are just 1 index":

```rs
struct Matrix(T![Function], T![Number], ...)
```

but with the optional commas, and needing to retain those for cursors' reasons, when wanting the third argument, youd need like `.5`.

I could be entirely wrong, and its not expected to work with the ast, and how its represented is irrelevant.

crates/css_ast/src/types/transform_list.rs
```diff
@@ -1,32 +1,75 @@
...
+		assert_parse!(TransformList, "rotate(45deg)");
+		assert_parse!(TransformList, "skew(1deg,2deg)");
+		assert_parse!(TransformList, "skewX(1deg)");
+		assert_parse!(TransformList, "skewY(1deg)");
+		assert_parse!(TransformList, "rotate(180deg)scale(2,3)");
+		assert_parse!(TransformList, "skewX(10deg)skewY(20deg)rotate(45deg)");
+		assert_parse!(TransformList, "scale(1.5)rotate(90deg)skew(15deg,30deg)");
+		assert_parse!(
+			TransformList,
+			"matrix(1,0,0,1,0,0)translate(1rem)translateX(1rem)translateY(1rem)scale(2)scale(1,2)scaleX(2)scaleY(2)rotate(45deg)skew(1deg,2deg)skewX(1deg)skewY(1deg)"
```
maraisr, line 66: Just a test to ensure everything can be chained. But question @keithamus, are whitespaces lost in tokens? Be kinda nice to be able to:

```diff
- assert_parse!(TransformList, "rotate(180deg)scale(2,3)");
+ assert_parse!(TransformList, "rotate(180deg) scale(2,3)");
```

Tests fail coz its missing a space.

crates/css_ast/src/types/transform/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+	Rotate: "rotate",
+	Skew: "skew",
+	SkewX: "skewx",
+	SkewY: "skewy",
+});
+
+// https://drafts.csswg.org/css-transforms-1/#two-d-transform-functions
+#[derive(Parse, ToCursors, Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize), serde())]
+pub enum TransformFunction {
```
maraisr: This is so convenient, just derive Parse. But can see us _not_ doing that, so individual functions can have simpler impls. And handling the parse steps here. Not sure if individual functions are ever parsed outside of just here, so we wont need the safety that say Matrix::parse_steps has, but checking that the function name is matrix. 

I guess what im trying to ask is, should each fn be able to be parsed on its own? Maybe not, and this enum here should own that logic, and thus simplify all the fns.

keithamus (approved):

PR #163, opened 2025-07-07
Derive to_cursors where we can instead of blanket impl

maraisr:


keithamus (approved):
This is a really nice tidy up!

PR #168, opened 2025-07-08
Enhances CI to also run doc builds

maraisr:
We have been failing CI for releasing the docs site, due to a broken `cargo doc`. What this PR does is;

1. Builds docs as part of the tests checks run—so we can see if they also break
2. Runs the production docs with `--keep-going --release`, so if there are still errors—still build the docs (`--keep-going`) them, as well as build them in release mode (`--release`)
3. Copied the same matrix from release into test, as arguably tests should pass on what we release 
4. Fixes up the docs so they actually pass green again
5. Similarly to how we do things on dotcom, adds a "result" check—which we can then mark as the required check, so we can add/remove matrix builds without syncing them in the UI.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -66,3 +66,10 @@ jobs:
         uses: Swatinem/rust-cache@v2
         # Ensure that docs can also build
       - run: cargo doc --all-features --no-deps
+
+  results:
+    runs-on: ubuntu-latest
+    needs: [build, docs]
+    steps:
+      - name: results
+        run: echo ${{ toJSON(needs.*.result) }}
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -40,8 +56,20 @@ jobs:
...
+        uses: Swatinem/rust-cache@v2
+        # Ensure that docs can also build
+      - run: cargo doc --all-features --no-deps
+
+  results:
+    runs-on: ubuntu-latest
+    needs: [build, docs]
+    steps:
+      - name: results
+        run: echo '${{ toJSON(needs.*.result) }}'
```
keithamus, line 75: Does this actually fail a build?

keithamus (approved):
Docs stuff LGTM. Feel less sure about the results check but happy to merge and see how it goes.

.github/workflows/test.yml
maraisr: If one of the prior "needs" steps fail, then this check will never run. So wont report, and wont allow a check to pass a PR.

PR #171, opened 2025-07-09
(css_ast): enables all transition properties

### LINKED ISSUE #160: Todo: <single-transition>
[`<single-transition>`](https://github.com/csskit/csskit/blob/main/crates/css_ast/src/types/single_transition.rs) is a stub, but it's needed for the `transition` property.

https://drafts.csswg.org/css-transitions-1/#single-transition
```
<single-transition> = [ none | <single-transition-property> ] || <time> || <easing-function> || <time> 
```

[~80% of pages use this. So very high popularity](https://chromestatus.com/metrics/css/timeline/popularity/150).

maraisr:
Implements both `<single-transition>` as well as `<single-transition-property>` which then meant we can uncomment all transition properties.

closes: https://github.com/csskit/csskit/issues/160

### DIFF
--- a/crates/css_ast/src/types/single_transition_property.rs
+++ b/crates/css_ast/src/types/single_transition_property.rs
@@ -1,32 +1,46 @@
 #![allow(warnings)]
 use css_lexer::{Cursor, SourceOffset};
 use css_parse::{CursorSink, Parse, Parser, Peek, Result as ParserResult, T, ToCursors};
+use csskit_derives::{Peek, ToCursors};
 
 // https://drafts.csswg.org/css-transitions-1/#single-transition-property
 // <single-transition-property> = all | <custom-ident>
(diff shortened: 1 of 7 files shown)


crates/css_ast/src/types/single_transition_property.rs
```diff
@@ -1,32 +1,46 @@
...
 }
 
 impl<'a> Parse<'a> for SingleTransitionProperty {
 	fn parse(p: &mut Parser<'a>) -> ParserResult<Self> {
-		todo!();
-	}
-}
+		let ident = p.parse::<T![Ident]>()?;
+		if p.eq_ignore_ascii_case(ident.into(), "all") {
+			return Ok(SingleTransitionProperty::All(ident));
```
keithamus, line 19: Wonder if this should use `keyword_set!(AllKeyword, "all")`?

keithamus (approved):

PR #172, opened 2025-07-09
(css_ast): refactors `::units::time` to peek/parse slightly better

maraisr:
There is a slight optimization here in that we don't need to `parse_str_lower(c)` and can rely on the dimension's peek steps, like checking the dimension_unit.


keithamus (approved):
I bet there are a few others like this. Good stuff.

PR #175, opened 2025-07-10
(css_ast): enables `<content-list>`

### LINKED ISSUE #154: Todo: <content-list>
[`<content-list>`](https://github.com/csskit/csskit/blob/main/crates/css_ast/src/types/content_list.rs) is a stub, but it's needed for `string-set`, `bookmark-label` and `content` properties.

https://drafts.csswg.org/css-content-3/#content-values
```
<content-list> = [ <string> | <image> | <attr()> | contents | <quote> | <leader()> | <target> | <string()> | <content()> | <counter> ]+
```

This is a rule with many variants, so it may be possible to half-finish this, such that we can enable properties like `content` and have them pass the test suite, despite not implementing some variants (I imagine `leader()` is not widely used, but `<string>` will be, `<attr()>` less so, etc).

[`content` is the most popular, and is om 80% of pages. So extremely high popularity](https://chromestatus.com/metrics/css/timeline/popularity/74).

maraisr:
To enable `<content-list>` we needed to enable a bunch of other types, like `<leader-type>`, `<leader()>`, `<content()>`, `<counter>` and more. 

I also added blanket implementation for `Vec<\'_, T>` that implements `Parse` too, which unlocks being able to derive Parse for things like `struct(Vec<'a, Foobar>)`.

I can see a second pass through that could enable deriving Parse for an enum of functions, and for something like `Option<T>`. As well as a bit of a tidy up to make use of `Vec::parse`.

closes: https://github.com/csskit/csskit/issues/154


keithamus (approved):
Looks great! Let's merge and iterate.

PR #178, opened 2025-07-11
(css_ast): run cargo fmt

maraisr:
`main` is broken right now, simply ran `cargo fmt` to fix this build. Proved this will properly fix CI, in this pr: https://github.com/csskit/csskit/pull/179


keithamus (approved):
Oops my bad

PR #181, opened 2025-07-11
Introduces Optionals![A, B, ...] and uses them in a bunch of places

maraisr:
This PR adds a new `Optionals![]` macro, to help with parsing types like `[ foo || bar ]`. And then goes and uses that in a few places like `<grid-line>`, and `<single-transition>`.


crates/css_ast/src/types/grid_line.rs
```diff
@@ -12,51 +12,39 @@ keyword_set!(GridLineKeywords { Auto: "auto", Span: "span" });
...
+	Span(GridLineKeywords, Option<PositiveNonZeroInt>, Option<T![Ident]>),
 	Area(T![Ident]),
 	Placement(T![Number], Option<T![Ident]>),
 }
 
+#[derive(ToSpan, Peek, ToCursors, Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize), serde())]
+pub struct PositiveNonZeroInt(pub T![Number]);
+
+impl<'a> Parse<'a> for PositiveNonZeroInt {
```
keithamus: We should move this into `types` - it can live alongside the other types. 

It also might be worthwhile wrapping `CSSInt` instead of `T![Number]` - `CSSInt` already checks for `.is_int()`.

crates/css_parse/src/macros/optionals.rs
```diff
@@ -0,0 +1,137 @@
...
+	use crate::test_helpers::*;
+	use crate::token_macros::*;
+
+	#[test]
+	fn size_test() {
+		assert_eq!(std::mem::size_of::<Optionals2<Ident, Number>>(), 32);
+	}
+
+	#[test]
+	fn test_writes() {
```
keithamus, line 135: Shall we also add some Span tests? Might be useful to ensure parsing returns when it sees something that isn't one of the types.

crates/css_parse/src/macros/optionals.rs
```diff
@@ -0,0 +1,137 @@
...
+		type CaseA = Optionals![Number, Ident];
+		assert_parse!(CaseA, "123 foo", Optionals2(Some(_), Some(_)));
+		assert_parse!(CaseA, "foo 123", "123 foo", Optionals2(Some(_), Some(_)));
+		assert_parse!(CaseA, "123", Optionals2(Some(_), None));
+		assert_parse!(CaseA, "foo", Optionals2(None, Some(_)));
```
keithamus, line 139: Would be great to test some of the other Optionals, as in the (A,B,C) or (A,B,C,D) or (A,B,C,D,E) variants.

keithamus (commented):
So far so good! Couple of small tweaks:

keithamus (approved):
Great stuff!

PR #224, opened 2025-07-31
tidy up github actions

keithamus:
This change cleans up all of the github actions to:

- Drop the `mise run install` step and use Mise to manage extra cargo deps. This keeps things much simpler for executions, and also managing the versions of these tools.
- Use Mise to manage the install dependencies. This keeps us consistent across all actions, and makes it much easier to manage dependencies.
- Use git shas for actions versioning, this keeps us more secure as tags cannot change from under us.
- Tidy up all caching, as some caches were improper.


keithamus:
Test time on win32-arm64 was ~5m, down now to ~3m, darwin-* from ~2m30s to ~1m30s, linux-* from ~1m30s to ~1m.

PR #429, opened 2025-09-08
css_lexer: mark a bunch of functions const

keithamus:
Silly, I know, but this seems to give us a ~5% improvement in parse times on my machine. Let's see what CI thinks.


keithamus:
Unsurprisingly, little difference in CI. I don't think the 5% I see locally is truly representative but also some of these consts will be convenient for constexpr stuff so I am merging anyway.

PR #532, opened 2025-10-06
Regenerate css_ast/src/values from csswg drafts

keithamus:
Updates to the latest specs


prettyGrace (commented):
Verify and update all components and remove solve all issues

prettyGrace (commented):
Approved

PR #582, opened 2025-10-20
Fix link to csskit.rs in README.

jnjaeschke:


keithamus:
Nice catch! Thank you!

keithamus:
Hmm... silly me; our CI fails because the check result tries to post a review comment but lacks permissions as it comes from a fork.

PR #799, opened 2026-01-07
css_ast: enables `<display>`

maraisr:
This PR implements the `DisplayStyleValue`, and dependant display types. We did need to expect the proc_macros to support a Combinator of Options, so did that in this diff as well.


crates/css_ast/src/types/display_listitem.rs
```diff
@@ -0,0 +1,102 @@
...
+		assert_parse!(CssAtomSet::ATOMS, DisplayListitem, "flow list-item");
+		assert_parse!(CssAtomSet::ATOMS, DisplayListitem, "flow-root list-item");
+		assert_parse!(CssAtomSet::ATOMS, DisplayListitem, "block flow list-item");
+		assert_parse!(CssAtomSet::ATOMS, DisplayListitem, "inline flow list-item");
+		assert_parse!(CssAtomSet::ATOMS, DisplayListitem, "block flow-root list-item");
+		assert_parse!(CssAtomSet::ATOMS, DisplayListitem, "inline flow-root list-item");
+		assert_parse!(CssAtomSet::ATOMS, DisplayListitem, "list-item block");
+		assert_parse!(CssAtomSet::ATOMS, DisplayListitem, "list-item flow");
+		assert_parse!(CssAtomSet::ATOMS, DisplayListitem, "list-item block flow");
+		assert_parse!(CssAtomSet::ATOMS, DisplayListitem, "flow list-item block");
```
keithamus, line 76: These aren't right. These should be errors. The `&&` semantics mean "in this order". 


```suggestion
		assert_parse_error!(CssAtomSet::ATOMS, DisplayListitem, "list-item block");
		assert_parse_error!(CssAtomSet::ATOMS, DisplayListitem, "list-item flow");
		assert_parse_error!(CssAtomSet::ATOMS, DisplayListitem, "list-item block flow");
		assert_parse_error!(CssAtomSet::ATOMS, DisplayListitem, "flow list-item block");
```

crates/css_ast/src/types/display_listitem.rs
```diff
@@ -0,0 +1,102 @@
...
+
+		let list_item = match list_item {
+			Some(li) if p.equals_atom(li.into(), &CssAtomSet::ListItem) => li,
+			Some(li) => Err(Diagnostic::new(li.into(), Diagnostic::unexpected_ident))?,
+			None => Err(Diagnostic::new(p.next(), Diagnostic::unexpected))?,
+		};
+
+		Ok(Self { outside, inside, list_item })
+	}
+}
```
keithamus, line 38: `&&` requires an order so `parse_optionals!` isn't correct here. Each one needs to be parsed in order, and the `list-item` is required.

I think the default `Parse` derive would give correct semantics:


```suggestion
#[derive(Parse, Peek, ToSpan, ToCursors, SemanticEq, Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
#[cfg_attr(feature = "serde", derive(serde::Serialize), serde())]
#[cfg_attr(feature = "visitable", derive(csskit_derives::Visitable), visit(children))]
pub struct DisplayListitem {
	pub outside: Option<DisplayOutside>,
	pub inside: Option<DisplayListitemInside>,
	pub list_item: T![Ident],
}
```

crates/css_ast/src/types/display_listitem.rs
```diff
@@ -0,0 +1,102 @@
...
+		assert_parse_error!(CssAtomSet::ATOMS, DisplayListitem, "block");
+		assert_parse_error!(CssAtomSet::ATOMS, DisplayListitem, "inline");
+		assert_parse_error!(CssAtomSet::ATOMS, DisplayListitem, "flow");
+		assert_parse_error!(CssAtomSet::ATOMS, DisplayListitem, "flow-root");
+		assert_parse_error!(CssAtomSet::ATOMS, DisplayListitem, "block flow");
+		assert_parse_error!(CssAtomSet::ATOMS, DisplayListitem, "inline flow-root");
+	}
+
+	#[test]
+	fn test_errors() {
```
keithamus: Ideally errors should be in their own test function.


```suggestion
	}

	#[test]
	fn test_errors() {
		// These tests verify that `list-item` is required (the `&&` semantics).
		// Without the explicit check for `list_item.is_some()`, these would incorrectly parse
		// because `parse_optionals!` only requires at least one component (the `||` semantics).
		assert_parse_error!(CssAtomSet::ATOMS, DisplayListitem, "block");
		assert_parse_error!(CssAtomSet::ATOMS, DisplayListitem, "inline");
		assert_parse_error!(CssAtomSet::ATOMS, DisplayListitem, "flow");
		assert_parse_error!(CssAtomSet::ATOMS, DisplayListitem, "flow-root");
		assert_parse_error!(CssAtomSet::ATOMS, DisplayListitem, "block flow");
		assert_parse_error!(CssAtomSet::ATOMS, DisplayListitem, "inline flow-root");
```

crates/csskit_proc_macro/src/generate.rs
```diff
@@ -99,14 +99,28 @@ impl ToFieldName for Def {
...
+					ds.iter().fold(String::new(), |str, d| match d {
+						Def::Type(d) => {
+							format!("{}{}", str, d.to_variant_name(0))
+						}
+						_ => {
+							dbg!("TODO variant name for Combinator() of Options", d);
+							todo!("variant name")
+						}
+					})
+				)
```
keithamus, line 114: As a nice follow up (not needed for this PR) it might be useful to take the generated name (`DisplayInsideDisplayOutside`) and map it via a toml file into a nicer name. In this case this name is _probably okay_ but `DisplayInsideOutside` or `DisplayMultiKeyword` or some other such name would probably make more sense, and while I think this is a reasonable heuristic which might get us far, I also worry that some names will end up extremely burdensome, for example `PositionVisibilityStyleValue` has `always | [ anchors-valid || anchors-visible || no-overflow ]` which would end up being `AnchorsValidAnchorsVisibileNoOverflow` which is a mouthful when it could be `PositionVisibilityKeyword` or so.

This will matter because each of these generated values is mapped to a node ID which makes it selectable via rule sheets, and the names should be clear and self explanatory there.

keithamus (changes requested):
Really excellent work! This will have a large knock-on impact for parsing a lot of new values, plus `display` itself is an incredibly popular property so we really need to get it parsing!

There are a few things that need to change, but mostly small, plus some commentary for follow up work:

crates/css_ast/src/types/display_listitem.rs
maraisr: Hmmmm, my understanding was that `&&` means in any order. Have you got a link to this so I can read further?

crates/css_ast/src/types/display_listitem.rs
maraisr: I do check that list-item is required, I'm just leveraging the `pare_optionals!` so I didn't need to implement the same thing. Since both of its members are `?`, its basically a `||`.

I think the other comment on `&&` could prove me wrong here, waiting for a link.

crates/css_ast/src/types/display_listitem.rs
maraisr: Agree! Thought I did that, ill move it.

crates/css_ast/src/types/display_listitem.rs
keithamus: https://drafts.csswg.org/css-values-4/#component-combinators

I am wrong! 

> A double ampersand (&&) separates two or more components, all of which must occur, in any order. 

In this I think we might be able to derive parse but add the `#[parse(all_must_occur)]`

```suggestion
#[derive(Parse, Peek, ToSpan, ToCursors, SemanticEq, Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
#[cfg_attr(feature = "serde", derive(serde::Serialize), serde())]
#[cfg_attr(feature = "visitable", derive(csskit_derives::Visitable), visit(children))]
#[parse(all_must_occur)]
pub struct DisplayListitem {
	pub outside: Option<DisplayOutside>,
	pub inside: Option<DisplayListitemInside>,
	pub list_item: T![Ident],
}
```

keithamus (approved):
Thanks for double checking me on the `&&`, you were totally right!

In that case the comments here are mostly nits, so this can be merged unless you want to follow up and address e.g. the test ordering and attempt derive?

crates/css_ast/src/types/display_listitem.rs
maraisr: This is so cool! `#[parse(all_must_occur)]` is so nice. Love finding these new things!!

I did try that, and looks like it doesn't handle `Option<T>` "properly" yet, I want to dive into that in a follow up.

crates/css_ast/src/types/display_listitem.rs
keithamus: I misinterpreted, you are correct:

[drafts.csswg.org/css-values-4#component-combinators](https://drafts.csswg.org/css-values-4/#component-combinators)

>    A double ampersand (&&) separates two or more components, all of which must occur, in any order.

Issue #777, opened 2026-01-02
Todo: Minify lengths


keithamus:
Lengths should be written out as their shortest unit, e.g. `96px` can be written as `1in`, `0px` can be written as `0`.

https://github.com/csskit/csskit/issues/774 needs to be resolved first.

yisibl:
> 0px can be written as 0

Historically, the `0px` unit cannot be omitted in certain properties or functions, which requires some investigation.
- https://github.com/cssnano/cssnano/issues/768
- https://github.com/sass/sass/issues/649#issuecomment-447033060

If possible, it would be best to provide an option(`ignoreProperties`/`ignoreFunctions `) allowing users to configure this list themselves.
- https://stylelint.io/user-guide/rules/length-zero-no-unit/
- https://ota-meshi.github.io/eslint-plugin-css/rules/no-length-zero-unit.html

yisibl:
Additionally, do not omit the unit 0 in `@supports`; it should be output as-is.

## input
```css
@supports (width: 0px;) {
  .foo {}
}
```

## ❌output
```css
@supports(width:0;){.foo{}}
```

See https://csskit.rs/playground/?code=cyguLSjILyopVtAoz0wpybBSMCiosNZUqOZSUNBLy89XqK7lquXickAoy0st1y0oyi9ILSqpRFatoJCUn1IJZYK5icnZ6UX5pXkpVgpJOYnJ2dZgqVquWgA%3D&format=minify

yisibl:
Then, I hope we avoid touching `0%` in the short term. Historically, many CSS minifiers have arrogantly converted `0%` to `0`, which is unsafe. https://github.com/cssnano/cssnano/issues/229#issuecomment-3771174416

I have high expectations for the `csskit` project. I hope it will be safe and standards-compliant, establishing itself as the community's standard implementation.

yisibl:
Please reopen this issue. I believe we still have a lot of work to do.

This also requires further consideration.

```css
.foo {
  flex: 0 1 0%;
  flex: 0 1 0px;
  filter: blur(0px);
  /* and more ... */
}
```

keithamus:
@yisibl I raised https://github.com/csskit/csskit/issues/820. We will come up with an elegant fix that reasons about the properties.

I don't intend to minify %s until we can work out how they resolve, per property.



> Please reopen this issue. I believe we still have a lot of work to do.
> 
> This also requires further consideration.
> 
> .foo {
>   flex: 0 1 0%;
[...]

PR #821, opened 2026-01-21
css_ast/css_transform: Only reduce lengths to unitless zero when applicable

### LINKED ISSUE #820: Lengths cannot always be minified to 0
From https://github.com/csskit/csskit/issues/777#issuecomment-3771092758

> @yisibl
> > 0px can be written as 0
> 
> Historically, the `0px` unit cannot be omitted in certain properties or functions, which requires some investigation.
> 
>     * [line-height: 0px != line-height: 0 cssnano/cssnano#768](https://github.com/cssnano/cssnano/issues/768)
> 
>     * [Omit units for zero values sass/sass#649 (comment)](https://github.com/sass/sass/issues/649#issuecomment-447033060)
> 
> 
> If possible, it would be best to provide an option(`ignoreProperties`/`ignoreFunctions `) allowing users to configure this list themselves.
> 
>     * [stylelint.io/user-guide/rules/length-zero-no-unit](https://stylelint.io/user-guide/rules/length-zero-no-unit/)
> 
>     * [ota-meshi.github.io/eslint-plugin-css/rules/no-length-zero-unit.html](https://ota-meshi.github.io/eslint-plugin-css/rules/no-length-zero-unit.html)

I think we should avoid the `ignoreProperties`/`ignoreFunctions`. We should ensure that we're only translating length to Zero directly when we know Zero is a valid value. This can be driven by the AST, so it should not need to be configurable.

keithamus:
In #777 @yisibl pointed out a bug in ReduceLengths (#815) in that unitless zero
is not always semantically equivalent to `0px`. This is because:

- Some values, e.g. `line-height` and `tab-size` have distinct `<number>` and
  `<length>` types. These types are distinct because they have semantic meaning
	(line-height uses number as a multiplier, so `1` is "1x font-size", tab-size
	uses number as "the number of spaces"). While for a single property a
	line-height of `0` vs `0px` might be _visually_ equivalent, this can cause
	issues with animation.
- Calc's system for resolving the type requires units to be equivalent, so
	`calc(100px - 0px)` returns a Length, while `calc(100px - 0)` returns a Number
	which will be invalid in some contexts.
- For unknown properties (or custom ones) we cannot make the determination
	whether or not they'll be used in calc, or whether or not this is a property
	we're not-yet parsing.

This was separated out into #820.

Because of these gotchas, unitless zero is not _always_ equivalent to `0px`.
Therefore it is not safe for us to always convert `0px` to unitless zero.
However, we can reason about this using metadata.

This change introduces a new UnitlessZeroResolves metadata bit, which can be
used to determine how a declaration might reason about unitless zero in an
abstract way (without looking up property names). The metadata system propagates
through nodes, so later on when we parse calc/custom properties better, we can
set the UnitlessZeroResolves::Number flag and it should retain this new
behaviour.

ReduceLengths has now been extended so that when it visits a declaration it will
set a boolean based on this value, meaning traversing into the declaration, it
can be aware as to whether or not it should coerce these values.

This allows us to continue to reduce `0px` to `0` when they are semantically
equivalent, while also ensuring we're only doing so when they actually are.

This fixes #820.


yisibl:
Nice work!

PR #848, opened 2026-01-30
website: fix colors in dark mode

### LINKED ISSUE #847: Playground colours are broken in dark mode
/cc @yisibl 

The radios are rendered in a light text colour regardless of dark mode.

Image

Image

keithamus:
Fixes #847

### DIFF
--- a/website/css/playground.css
+++ b/website/css/playground.css
@@ -2,6 +2,10 @@ main {
 	height: 100vh;
 }
 
+example-picker, metric-observer {
+	color: var(--text-1);
+}
+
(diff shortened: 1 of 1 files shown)


website/css/playground.css
```diff
@@ -2,6 +2,10 @@ main {
 	height: 100vh;
 }
 
+example-picker, metric-observer {
+	color: var(--text-1);
+}
+
```
yisibl, line 8: If we don't need to be compatible with older browsers, we can try `light-dark()`.

Issue #157, opened 2025-07-06
Todo: <gap-rule-list>


keithamus:
[`<gap-rule-list>`](https://github.com/csskit/csskit/blob/main/crates/css_ast/src/types/gap_rule_list.rs) is a stub, but it's needed for both `row-rule` and `column-rule` properties.

https://drafts.csswg.org/css-gaps-1/#typedef-gap-rule-list
```
<gap-rule-list> = <gap-rule-or-repeat>#
```

[~2.5% of pages use this. So low-ish popularity](https://chromestatus.com/metrics/css/timeline/popularity/525).

cijiugechu:
Please assign it to me, I’d like to work on this issue.

keithamus:
This can now be closed. Thanks for working on this @cijiugechu! I hope we get more wonderful contributions from you 🫶

PR #878, opened 2026-02-12
css_ast: Implement gap rule list

cijiugechu:
Related issue: #157


crates/css_ast/src/types/gap_auto_rule_list.rs
```diff
@@ -1,14 +1,225 @@
...
+#[derive(csskit_derives::NodeWithMetadata)]
+pub enum GapAutoRuleListItem<'a> {
+	GapAutoRepeatRule(GapAutoRepeatRule<'a>),
+	GapRuleOrRepeat(GapRuleOrRepeat<'a>),
+}
+
+// <gap-auto-repeat-rule> = repeat( auto , <gap-rule># )
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize), serde())]
+#[cfg_attr(feature = "visitable", derive(csskit_derives::Visitable), visit(self))]
```
keithamus: Same as above; `visit(self)` is generally for opaque types, so this should be just `visit()`:

```suggestion
#[cfg_attr(feature = "visitable", derive(csskit_derives::Visitable), visit())]
```

Or if this _should_ be opaque then drop the `visit(skip)`s below as they're implied.

crates/css_ast/src/types/gap_auto_rule_list.rs
```diff
@@ -1,14 +1,225 @@
...
 // https://drafts.csswg.org/css-gaps-1/#typedef-gap-auto-rule-list
 // <gap-auto-rule-list> = <gap-rule-or-repeat>#? , <gap-auto-repeat-rule> , <gap-rule-or-repeat>#?
-pub type GapAutoRuleList = Todo;
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize), serde())]
+#[cfg_attr(feature = "visitable", derive(csskit_derives::Visitable), visit(self))]
+#[derive(csskit_derives::NodeWithMetadata)]
+pub struct GapAutoRuleList<'a> {
+	pub items: Vec<'a, (GapAutoRuleListItem<'a>, Option<T![,]>)>,
+}
```
keithamus: If this is a newtype this _probably_ wants to be simply:


```suggestion
pub struct GapAutoRuleList<'a>(CommaSeparated<'a, GapAutoRuleListItem<'a>>);
```

crates/css_ast/src/types/gap_auto_rule_list.rs
```diff
@@ -1,14 +1,225 @@
...
-
-use crate::Todo;
+use crate::{GapRule, GapRuleOrRepeat};
 
 // https://drafts.csswg.org/css-gaps-1/#typedef-gap-auto-rule-list
 // <gap-auto-rule-list> = <gap-rule-or-repeat>#? , <gap-auto-repeat-rule> , <gap-rule-or-repeat>#?
-pub type GapAutoRuleList = Todo;
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize), serde())]
+#[cfg_attr(feature = "visitable", derive(csskit_derives::Visitable), visit(self))]
```
keithamus: If this is just `visit(self)` then you won't need the `visit(skip)` attributes below, or you want to traverse children (probably preferred) so this should be just `visit()`:

```suggestion
#[cfg_attr(feature = "visitable", derive(csskit_derives::Visitable), visit())]
```

crates/css_ast/src/types/gap_auto_rule_list.rs
```diff
@@ -1,14 +1,225 @@
...
+	loop {
+		let item = p.parse::<T>()?;
+		let comma = p.parse_if_peek::<T![,]>()?;
+		items.push((item, comma));
+		if items.last().and_then(|(_, c)| *c).is_none() {
+			break;
+		}
+	}
+	Ok(items)
+}
```
keithamus: Likely this can be dropped if you're using `CommaSeparated<T>`

crates/css_ast/src/types/gap_auto_rule_list.rs
```diff
@@ -1,14 +1,225 @@
...
+	#[cfg_attr(feature = "visitable", visit(skip))]
+	pub name: T![Function],
+	#[cfg_attr(feature = "visitable", visit(skip))]
+	pub count: T![Ident],
+	#[cfg_attr(feature = "visitable", visit(skip))]
+	pub comma: T![,],
+	pub rules: Vec<'a, (GapRule<'a>, Option<T![,]>)>,
+	#[cfg_attr(feature = "visitable", visit(skip))]
+	pub close: T![')'],
+}
```
keithamus: We actually have `RepeatFunction` already: https://github.com/csskit/csskit/blob/main/crates/css_ast/src/functions/repeat_function.rs

I wonder if we should extend the `GapRule` into `RepeatFunctionParams`?

crates/css_ast/src/types/gap_auto_rule_list.rs
```diff
@@ -1,14 +1,225 @@
...
+		let last_span = last.1.map_or_else(|| last.0.to_span(), |comma| comma.to_span());
+		first + last_span
+	}
+}
+
+impl<'a> css_parse::ToSpan for GapAutoRepeatRule<'a> {
+	fn to_span(&self) -> css_parse::Span {
+		self.name.to_span() + self.close.to_span()
+	}
+}
```
keithamus: These can all likely be swapped out for `derive(ToCursors, ToSpan)`

crates/css_ast/src/types/gap_auto_rule_list.rs
```diff
@@ -1,14 +1,225 @@
...
+	fn semantic_eq(&self, other: &Self) -> bool {
+		self == other
+	}
+}
+
+impl<'a> css_parse::SemanticEq for GapAutoRepeatRule<'a> {
+	fn semantic_eq(&self, other: &Self) -> bool {
+		self == other
+	}
+}
```
keithamus: These don't look right. SemanticEq needs to check semantic children, as partial equality will look at Cursor offsets. You probably just want to `derive(SemanticEq)` but if you would rather hand roll it you'll need to call `self.foo.semantic_eq(other.foo)` for each property of the struct. I would recommend derive.

crates/css_ast/src/types/gap_rule_list.rs
```diff
@@ -1,14 +1,274 @@
...
 // https://drafts.csswg.org/css-gaps-1/#typedef-gap-rule-list
 // <gap-rule-list> = <gap-rule-or-repeat>#
-pub type GapRuleList = Todo;
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize), serde())]
+#[cfg_attr(feature = "visitable", derive(csskit_derives::Visitable), visit(self))]
+#[derive(csskit_derives::NodeWithMetadata)]
+pub struct GapRuleList<'a> {
+	pub items: Vec<'a, (GapRuleOrRepeat<'a>, Option<T![,]>)>,
+}
```
keithamus: Again here you probably want:


```suggestion
pub struct GapRuleList<'a>(CommaSeparated<'a, GapRuleOrRepeat<'a>>);
```

crates/css_ast/src/types/gap_auto_rule_list.rs
```diff
@@ -1,14 +1,225 @@
...
+		let count = p.parse::<T![Ident]>()?;
+		if !p.equals_atom(count.into(), &CssAtomSet::Auto) {
+			Err(Diagnostic::new(count.into(), Diagnostic::unexpected_ident))?
+		}
+		let comma = p.parse::<T![,]>()?;
+		let rules = parse_comma_separated::<_, GapRule<'a>>(p)?;
+		let close = p.parse::<T![')']>()?;
+		Ok(Self { name, count, comma, rules, close })
+	}
+}
```
keithamus: With `CommaSeaprated` you could _probably_ `derive(Parse)` on `GapAutoRepeatRule` - if you add the `#[atom]` annotations to the Function/Ident members.

crates/css_ast/src/types/gap_rule_list.rs
```diff
@@ -1,14 +1,274 @@
...
+
+// <gap-rule> = <line-width> || <line-style> || <color>
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize), serde())]
+#[cfg_attr(feature = "visitable", derive(csskit_derives::Visitable), visit(self))]
+#[derive(csskit_derives::NodeWithMetadata)]
+pub struct GapRule<'a> {
+	pub items: Vec<'a, GapRuleItem>,
+}
```
keithamus: Hmm I don't think a `Vec` is _quite_ right here if this is a set of optionals. This is more something like:


```suggestion
pub type GapRule<'a> = Optionals3<LineWidth, LineStyle, Color>
```

At which point we can drop `GapRuleItem` and likely just `derive(Parse)`.

crates/css_ast/src/values/gaps/mod.rs
```diff
@@ -37,7 +37,7 @@ use impls::*;
 #[cfg_attr(feature = "css_feature_data", derive(ToCSSFeature), css_feature("css.properties.column-rule"))]
 #[cfg_attr(feature = "visitable", derive(Visitable), visit)]
 #[derive(csskit_derives::NodeWithMetadata)]
-pub enum ColumnRuleStyleValue {}
+pub enum ColumnRuleStyleValue<'a> {}
```
keithamus: So all `values/*/mod.rs` files are generated. If you're modifying them it's likely you'll need to instead modify `css_spec_generator` and/or `css_value_definition_parser`. You've already done `css_value_definition_parser` which is good.

`css_spec_generator` determines which derives each property gets `derive(Parse)`. If you _really_ need to manually Parse (which, for the most part, you likely won't) then you need to add this property to the list in a toml file: https://github.com/csskit/csskit/blob/main/crates/csskit_spec_generator/manual_parse_properties.toml

crates/css_ast/src/values/gaps/impls.rs
```diff
@@ -1 +1,77 @@
...
+	let checkpoint = p.checkpoint();
+	if let Ok(value) = p.parse::<GapRuleList<'a>>() {
+		return Ok(from_rule(value));
+	}
+	p.rewind(checkpoint);
+	if let Ok(value) = p.parse::<GapAutoRuleList<'a>>() {
+		return Ok(from_auto(value));
+	}
+	Err(Diagnostic::new(start, Diagnostic::unexpected))
+}
```
keithamus: I'm a little surprised by this, but I think this is perhaps due to the strict adherance to the CSS specified types of `GapRuleList` and `GapAutoRuleList`.

I don't think we need to specify these types _exactly_ (if we did we could probably auto generate a whole bunch), instead we should allow for some interpretation, with the view that other layers (the linter, or maybe some parser modes) can catch improper uses. To that end we can probably flatten `GapRuleList` and `GapAutoRuleList` into 1, and maybe flatten them into `RepeatFunction` too.

keithamus (commented):
Great work on this so far! Amazing to see!

I've got some early feedback which hopefully you'll find useful. I think the feedback largely applies to both `gap_rule_list`/`gap_auto_rule_list` so I haven't repeated it for each.

Generally speaking, I think you should look at `RepeatFunction` and see if we can extend into that, here.

cijiugechu:
> Great work on this so far! Amazing to see!
> 
> I've got some early feedback which hopefully you'll find useful. I think the feedback largely applies to both `gap_rule_list`/`gap_auto_rule_list` so I haven't repeated it for each.
> 
> Generally speaking, I think you should look at `RepeatFunction` and see if we can extend into that, here.

Thanks a lot for the thorough review — really appreciate the time and the clear suggestions!
I have applied patches in latest revision.

crates/css_ast/src/types/gap_auto_rule_list.rs
```diff
@@ -1,14 +1,50 @@
 #![allow(unused)]
-use super::prelude::*;
-
-use crate::Todo;
+use crate::{GapRepeatRule, GapRuleList, GapRuleOrRepeat};
 
 // https://drafts.csswg.org/css-gaps-1/#typedef-gap-auto-rule-list
 // <gap-auto-rule-list> = <gap-rule-or-repeat>#? , <gap-auto-repeat-rule> , <gap-rule-or-repeat>#?
-pub type GapAutoRuleList = Todo;
+//
```
keithamus: Nit: 
```suggestion

```

crates/css_ast/src/types/gap_rule_list.rs
```diff
@@ -1,14 +1,107 @@
...
+#[derive(Parse, Peek, ToCursors, ToSpan, SemanticEq, Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize), serde())]
+#[cfg_attr(feature = "visitable", derive(csskit_derives::Visitable), visit(self))]
+#[derive(csskit_derives::NodeWithMetadata)]
+pub struct GapRuleList<'a>(pub CommaSeparated<'a, GapRuleOrRepeat<'a>>);
+
+// <gap-rule-or-repeat> = <gap-rule> | <gap-repeat-rule>
+#[derive(Parse, Peek, ToCursors, ToSpan, SemanticEq, Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize), serde())]
+#[cfg_attr(feature = "visitable", derive(csskit_derives::Visitable), visit(self))]
```
keithamus: These still need to visit children I believe. Same for many others:

```suggestion
#[cfg_attr(feature = "visitable", derive(csskit_derives::Visitable), visit)]
```

The general rule is "if a type just holds cursors, it should be `visit(self)`. If it is just a delegate type (like a generic) it should be `visit(children)`, otherwise `visit`."

crates/css_ast/src/types/gap_rule_list.rs
```diff
@@ -1,14 +1,107 @@
...
+		if !p.equals_atom(name.into(), &CssAtomSet::Repeat) {
+			Err(Diagnostic::new(name.into(), Diagnostic::unexpected_ident))?
+		}
+		let count = p.parse::<AutoOr<PositiveNonZeroInt>>()?;
+		let comma = p.parse::<T![,]>()?;
+		let rules = p.parse::<CommaSeparated<'a, GapRule>>()?;
+		let close = p.parse::<T![')']>()?;
+		Ok(Self { name, count, comma, rules, close })
+	}
+}
```
keithamus: This looks fairly straightforward (nothing surprising). I'm quite sure `GapRepeatRule` can `derive(Parse)` if you add `#[atom(CssAtomSet::Repeat)]` to the `name` field.

crates/css_ast/src/values/gaps/impls.rs
```diff
@@ -1 +1,57 @@
...
+
+impl<'a> Parse<'a> for RowRuleStyleValue<'a> {
+	fn parse<I>(p: &mut Parser<'a, I>) -> ParseResult<Self>
+	where
+		I: Iterator<Item = Cursor> + Clone,
+	{
+		p.parse::<GapRuleList<'a>>().map(Self::GapRuleList)
+	}
+}
```
keithamus: These aren't quite right I think. 

This will cause the `RowRuleStyleValue` to be:

```rust
enum RowRuleStyleValue {
	GapRuleList(GapRuleList<'a>),
  GapAutoRuleList(GapAutoRuleList<'a>),
}
```

But `Parse` will only ever return the `GapRuleList()` variant, rendering the other variant useless. 

There are two options here:

1. Play around with the syntax optimisation phase to flatten these (take a look at https://github.com/csskit/csskit/blob/main/crates/css_value_definition_parser/src/lib.rs#L356-L520). This will mean that `RowRuleStyleValue` will become a struct over `GapRuleList`. I think this is _probably_ want we want.
2. Figure out _after_ parse if `GapRuleList` fits the `auto` part by interrogating its members and see if it belongs in `GapRuleList` or `GapAutoRuleList` (probably not what we want).

crates/css_ast/src/types/gap_rule_list.rs
cijiugechu: Done

crates/css_ast/src/types/gap_auto_rule_list.rs
```diff
@@ -1,14 +1,49 @@
 #![allow(unused)]
```
keithamus: Is this file needed any more now? I suspect it isn't but perhaps this line meant no compile error. If it is needed we should be able to drop this:

```suggestion
```

crates/css_ast/src/values/gaps/impls.rs
```diff
@@ -1 +1,35 @@
...
+		assert_parse!(CssAtomSet::ATOMS, RuleStyleValue, "1px solid red, repeat(auto, 2px dashed green)");
+	}
+
+	#[test]
+	fn test_errors() {
+		assert_parse!(
+			CssAtomSet::ATOMS,
+			ColumnRuleStyleValue,
+			"repeat(auto, 1px solid red), repeat(auto, 2px dashed green)"
+		);
```
keithamus, line 17: This `assert_parse!` needs to be moved up to `test_writes()`

crates/css_ast/src/values/gaps/impls.rs
```diff
@@ -1 +1,35 @@
...
+		assert_parse_error!(CssAtomSet::ATOMS, ColumnRuleStyleValue, "repeat(auto,)");
+		assert_parse!(
+			CssAtomSet::ATOMS,
+			RowRuleStyleValue,
+			"repeat(auto, 1px solid red), repeat(auto, 2px dashed green)"
+		);
```
keithamus, line 23: This `assert_parse!` needs to be moved up to `test_writes()`

keithamus (commented):
This is in a really good place! I think there's a few minor issues however:

crates/css_ast/src/values/gaps/impls.rs
cijiugechu: Done

keithamus (approved):
Looking great! Let's get this merged in!

PR #1112, opened 2026-05-12
css_parse: Preserve semantic meaning when removing comments between tokens

### LINKED ISSUE #770: Comments maybe can't be replaced by whitespace in custom properties
Bit of an exotic example and very unlikely that authors are doing this but caught this because my own tool to remove comments and whitespace from CSS was also doing it wrong.

See: https://codepen.io/romainmenke/pen/jErbPLz

https://csskit.rs/playground/?code=S8ksU6jmUlDQ1U1KLLJSSNTX0tJPsuaq5eJyyE1NyUxUKC6pzEnVQJHWBOtIySwD8fTS8vPBfAWF5PycfKCi9KLU1DxroEgtVy0A&format=minify

```css
div {
  --bar: a/**/b;
}
```

Becomes:

```css
div{--bar:a b}
```

Which would not match a style query that wasn't transformed in the same way:

```css
@container style(--bar: a/**/b) {}
```

This could be problematic because there is no guarantee that all styles will be processed by tools (e.g. inline styles).

----

Nice to see all the progress here 🎉

kimjune01:
## Problem

The minifier was replacing comments with whitespace even when no whitespace existed in the original source. For custom properties like `--bar:a/**/b`, this changed the component values from `ab` to `a b`, breaking style query matching with `@container style(--bar:a/**/b)`.

## Fix

Skip comments entirely in compact mode without outputting anything. When a separator is needed between tokens, check the source text gap for actual whitespace - only add a separator if whitespace was present in the original. This preserves the semantic difference between `a/**/b` (no space) and `a /* comment */ b` (has space).

Adds comprehensive tests for:
- Comments without whitespace between tokens
- Multiple consecutive comments
- Comments with surrounding whitespace
- Style query matching after minification

Fixes #770

### DIFF
--- a/crates/css_parse/src/cursor_compact_write_sink.rs
+++ b/crates/css_parse/src/cursor_compact_write_sink.rs
@@ -213,4 +213,12 @@ mod test {
 		assert_format!("foo  ", "foo");
 		assert_format!("foo; ", "foo;");
 	}
+
+	#[test]
+	fn test_preserves_comment_absence_in_custom_properties() {
+		// Issue #770: Comments in custom properties should not be replaced by whitespace
+		// because `--bar: a/**/b` and `--bar: a b` are different component values
+		assert_format!("div{--bar:a/**/b}", "div{--bar:a/**/b}");
+		assert_format!("div { --bar: a/**/b }", "div{--bar:a/**/b}");
+	}
 }
--- a/crates/css_parse/src/cursor_compact_write_sink.rs
+++ b/crates/css_parse/src/cursor_compact_write_sink.rs
@@ -8,8 +8,8 @@ pub struct CursorCompactWriteSink<'a, T: SourceCursorSink<'a>> {
 	source_text: &'a str,
 	sink: T,
 	last_token: Option<Token>,
-	last_cursor: Option<Cursor>,
 	pending: Option<SourceCursor<'a>>,
+	pending_comment: Option<SourceCursor<'a>>,
 }
 
 const PENDING_KINDSET: KindSet = KindSet::new(&[Kind::Semicolon, Kind::Whitespace]);
@@ -23,10 +23,43 @@ const NO_WHITESPACE_AFTER_KINDSET: KindSet =
 
 impl<'a, T: SourceCursorSink<'a>> CursorCompactWriteSink<'a, T> {
 	pub fn new(source_text: &'a str, sink: T) -> Self {
-		Self { source_text, sink, last_token: None, last_cursor: None, pending: None }
+		Self { source_text, sink, last_token: None, pending: None, pending_comment: None }
 	}
 
 	fn write(&mut self, c: SourceCursor<'a>) {
+		if c == Kind::Eof {
+			return;
+		}
+
+		// Rule 1: Comment arrives — store or drop.
+		// Don't flush pending whitespace yet; it stays pending until a non-comment token
+		// decides whether the whitespace (or the comment) is needed.
+		if c == Kind::Comment {
+			// Determine the effective last token (pending whitespace counts as emitted whitespace
+			// for purposes of deciding whether a comment is needed)
+			let effective_last = if self.pending.is_some_and(|p| p == Kind::Whitespace) {
+				Some(Kind::Whitespace)
+			} else {
+				self.last_token.map(|t| t.kind())
+			};
+			if effective_last.is_some_and(|k| NO_WHITESPACE_AFTER_KINDSET.contains(k)) {
+				// Comment after a token that doesn't want whitespace after — drop
+				return;
+			}
+			if effective_last.is_some_and(|k| k == Kind::Whitespace) {
+				// Whitespace already separates — comment is redundant, drop
+				return;
+			}
+			if effective_last.is_none() {
+				// Comment at start of stream — drop
+				return;
+			}
+			// Store for potential replay as separator
+			self.pending_comment = Some(c);
+			return;
+		}
+
+		// Flush pending whitespace/semicolons now that we have a real (non-comment) token
 		let mut skip_separator_check = false;
 		if let Some(prev) = self.pending {
 			self.pending = None;
@@ -37,8 +70,9 @@ impl<'a, T: SourceCursorSink<'a>> CursorCompactWriteSink<'a, T> {
 			let is_redundant_whitespace = self.last_token.is_none()
 				|| prev == Kind::Whitespace && (c == NO_WHITESPACE_BEFORE_KINDSET || no_whitespace_after_last);
 			if !is_redundant_semi && !is_redundant_whitespace {
+				// Whitespace is being emitted, so any pending comment is redundant
+				self.pending_comment = None;
 				self.last_token = Some(prev.token());
-				self.last_cursor = Some(prev.cursor());
 				self.sink.append(prev.compact());
 			} else if no_whitespace_after_last {
 				// If we're skipping whitespace because the last token doesn't need whitespace after it,
@@ -46,10 +80,17 @@ impl<'a, T: SourceCursorSink<'a>> CursorCompactWriteSink<'a, T> {
 				skip_separator_check = true;
 			}
 		}
-		if c == Kind::Comment {
-			// Skip comments entirely in compact mode - don't output them
-			return;
+
+		// Rule 2: Non-comment arrives with stored comment — replay or discard
+		if let Some(comment) = self.pending_comment.take() {
+			if c != NO_WHITESPACE_BEFORE_KINDSET {
+				// Comment acts as separator — replay it
+				self.last_token = Some(comment.token());
+				self.sink.append(comment.compact());
+			}
+			// Otherwise discard the comment
 		}
+
 		if c == PENDING_KINDSET {
 			self.pending = Some(c);
 			return;
@@ -58,27 +99,9 @@ impl<'a, T: SourceCursorSink<'a>> CursorCompactWriteSink<'a, T> {
 			&& let Some(last) = self.last_token
 			&& last.needs_separator_for(c.token())
 		{
-			// Check if there was whitespace in the original source between last_cursor and c
-			// If there was only comments (no whitespace), don't add a separator
-			let had_whitespace = if let Some(last_cursor) = self.last_cursor {
-				let last_end = (last_cursor.offset().0 + last_cursor.len() as u32) as usize;
-				let curr_start = c.cursor().offset().0 as usize;
-				if last_end < curr_start && last_end < self.source_text.len() && curr_start <= self.source_text.len() {
-					let gap = &self.source_text[last_end..curr_start];
-					gap.chars().any(|ch| ch.is_ascii_whitespace())
-				} else {
-					true // If no gap or invalid range, assume we need separator
-				}
-			} else {
-				true // If no last cursor, assume we need separator
-			};
-
-			if had_whitespace {
-				self.sink.append(SourceCursor::SPACE);
-			}
+			self.sink.append(SourceCursor::SPACE);
 		}
 		self.last_token = Some(c.token());
-		self.last_cursor = Some(c.cursor());
 		// Normalize quotes
 		if c == Kind::String {
 			self.sink.append(c.with_quotes(QuoteStyle::Double).compact())
@@ -133,7 +156,7 @@ mod test {
 				let mut stream = CursorCompactWriteSink::new(source_text, &mut sink);
 				let lexer = Lexer::new(&EmptyAtomSet::ATOMS, source_text);
 				let mut parser = Parser::new(&bump, source_text, lexer);
-				parser.parse_entirely::<$struct>().output.unwrap().to_cursors(&mut stream);
+				parser.parse_entirely::<$struct>().with_trivia().to_cursors(&mut stream);
 			}
 			assert_eq!(sink, $after.trim());
 		};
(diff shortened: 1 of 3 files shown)


crates/css_parse/src/cursor_compact_write_sink.rs
```diff
@@ -213,4 +237,38 @@ mod test {
...
 		assert_format!("foo; ", "foo;");
 	}
+
+	#[test]
+	fn test_preserves_comment_absence_in_custom_properties() {
+		// Issue #770: Comments between tokens should not be replaced by whitespace
+		// When there's no whitespace between tokens, only comments, they should be removed
+		// without adding a separator (which would change the component values)
+		assert_format!("div{--bar:a/**/b}", "div{--bar:ab}");
+		assert_format!("div { --bar: a/**/b }", "div{--bar:ab}");
```
keithamus: These are wrong though, as this still changes the component value; `a b` is `<ident> <ident>` while `ab` is one `<ident>`. 

They shoul dbe:

```suggestion
		assert_format!("div{--bar:a/**/b}", "div{--bar:a/**/b}");
		assert_format!("div { --bar: a/**/b }", "div{--bar:a/**/b}");
```

crates/css_parse/src/cursor_compact_write_sink.rs
```diff
@@ -213,4 +237,38 @@ mod test {
...
+		// When there IS whitespace, it should be normalized to a single space
+		assert_format!("div{--bar:a /* comment */ b}", "div{--bar:a b}");
+	}
+
+	#[test]
+	fn test_edge_cases_multiple_comments() {
+		// Multiple consecutive comments without whitespace
+		assert_format!("div{--bar:a/**//**/b}", "div{--bar:ab}");
```
keithamus: Following that, this should be:

```suggestion
		assert_format!("div{--bar:a/**//**/b}", "div{--bar:a/**/b}");
```

crates/css_parse/src/cursor_compact_write_sink.rs
```diff
@@ -213,4 +237,38 @@ mod test {
...
+		assert_format!("div{--bar:a/*comment*/}", "div{--bar:a}");
+		// Whitespace, comment, whitespace
+		assert_format!("div{--bar:a  /**/  b}", "div{--bar:a b}");
+	}
+
+	#[test]
+	fn test_style_query_matching() {
+		// Real-world case from issue #770: style queries must match minified custom properties
+		// If we minify --bar:a/**/b to --bar:a b (with space), it won't match @container style(--bar:a/**/b)
+		assert_format!("@container style(--bar:a/**/b){}", "@container style(--bar:ab){}");
```
keithamus: And this should be:


```suggestion
		assert_format!("@container style(--bar:a/**/b){}", "@container style(--bar:a/**/b){}");
```

crates/css_parse/src/cursor_compact_write_sink.rs
```diff
@@ -52,9 +58,27 @@ impl<'a, T: SourceCursorSink<'a>> CursorCompactWriteSink<'a, T> {
...
+				let curr_start = c.cursor().offset().0 as usize;
+				if last_end < curr_start && last_end < self.source_text.len() && curr_start <= self.source_text.len() {
+					let gap = &self.source_text[last_end..curr_start];
+					gap.chars().any(|ch| ch.is_ascii_whitespace())
+				} else {
+					true // If no gap or invalid range, assume we need separator
+				}
+			} else {
+				true // If no last cursor, assume we need separator
+			};
```
keithamus: This doesn't look right to me. I don't think we care so much about whitespace, as much as we care that skipped tokens don't result in a situation where tokens need separating by whitespace (or equivalent).

I think storing the `last_cursor` is not necessary (we don't really need to check offsets, that's too much) but if we really need the cursor we should ditch `last_token` also; however I'm sure we don't.

I think the right solution here looks _something_ like:

1. If the current token is a comment, and if the last token is not `NO_WHITESPACE_AFTER_KINDSET`, then store the comment token in `last_token`. Otherwise return.
2. If the last token is a comment, and if the current token is not `NO_WHITESPACE_BEFORE_KINDSET`, replay the comment token.

This might cause more comments to be rendered, so we'd need a few more edge case tests that combine comments with different token types, especially whitespace, so e.g. `assert_format!("foo /**/bar", "foo bar");`, `assert_format!("foo{/**/bar}", "foo{bar}")`, and so on.

keithamus (commented):
Thanks so much @kimjune01 for tackling this one! I think this looks generally like the right place to solve this, but I think we need to look closer at the solution as I don't think this is in exactly the right direction.

Here's some suggestions:

kimjune01:
@keithamus I added these tests to cover what you mentioned:

  1. test_comments_as_separators_preserved — a/**/b stays a/**/b, consecutive comments dedupe
  2. test_comments_adjacent_to_structural_tokens_dropped — foo{/**/bar} → foo{bar}
  3. test_whitespace_already_separates_comment_dropped — foo /**/bar → foo bar
  4. test_comment_edge_cases — dimensions, stream start, after close paren

crates/css_parse/src/cursor_compact_write_sink.rs
```diff
@@ -213,4 +260,52 @@ mod test {
...
+
+	#[test]
+	fn test_comments_as_separators_preserved() {
+		// Comments acting as separators between tokens — preserved
+		assert_format!("div{--bar:a/**/b}", "div{--bar:a/**/b}");
+		assert_format!("div { --bar: a/**/b }", "div{--bar:a/**/b}");
+		// Multiple consecutive comments — only first kept (second is redundant after comment)
+		assert_format!("div{--bar:a/**//**/b}", "div{--bar:a/**/b}");
+		// Style queries — same rules
+		assert_format!("@container style(--bar:a/**/b){}", "@container style(--bar:a/**/b){}");
```
keithamus: I think we probably also want to ensure we always use an empty comment, so 

```suggestion
		assert_format!("@container style(--bar:a/*some comment text*/b){}", "@container style(--bar:a/**/b){}");
```

You maybe want to define SourceCursor::EMPTY_COMMENT, (which may need changing `Token::new_comment` to `pub(crate) const`.

crates/css_parse/src/cursor_compact_write_sink.rs
```diff
@@ -109,7 +156,7 @@ mod test {
 				let mut stream = CursorCompactWriteSink::new(source_text, &mut sink);
 				let lexer = Lexer::new(&EmptyAtomSet::ATOMS, source_text);
 				let mut parser = Parser::new(&bump, source_text, lexer);
-				parser.parse_entirely::<$struct>().output.unwrap().to_cursors(&mut stream);
+				parser.parse_entirely::<$struct>().with_trivia().to_cursors(&mut stream);
```
keithamus, line 138: While this passes these tests locally, we'll likely want to include trivia in all callsites that feed into CursorCompactWriteSink, so e.g. csskit_transform & csskit crates.

crates/css_parse/src/cursor_compact_write_sink.rs
```diff
@@ -44,6 +80,17 @@ impl<'a, T: SourceCursorSink<'a>> CursorCompactWriteSink<'a, T> {
...
 		}
+
+		// Rule 2: Non-comment arrives with stored comment — replay or discard
+		if let Some(comment) = self.pending_comment.take() {
+			if c != NO_WHITESPACE_BEFORE_KINDSET {
+				// Comment acts as separator — replay it
+				self.last_token = Some(comment.token());
+				self.sink.append(comment.compact());
+			}
+			// Otherwise discard the comment
```
keithamus: These comments aren't adding much value.

crates/css_parse/src/cursor_compact_write_sink.rs
```diff
@@ -22,10 +23,43 @@ const NO_WHITESPACE_AFTER_KINDSET: KindSet =
...
+				return;
+			}
+			if effective_last.is_none() {
+				// Comment at start of stream — drop
+				return;
+			}
+			// Store for potential replay as separator
+			self.pending_comment = Some(c);
+			return;
+		}
```
keithamus: I'm not sure why we're doing so much coercion/checking of Whitespace here. This feels like it could be cleaned up or simplified to something like:


```suggestion
                if c == Kind::Comment {
                        if self.pending.is_none()
                                && self.pending_comment.is_none()
                                && self.last_token.is_some_and(|t| t != NO_WHITESPACE_AFTER_KINDSET) {
                                self.pending_comment = Some(c);
                        }
                        return;
                }

```

crates/css_parse/src/cursor_compact_write_sink.rs
```diff
@@ -44,6 +80,17 @@ impl<'a, T: SourceCursorSink<'a>> CursorCompactWriteSink<'a, T> {
 				skip_separator_check = true;
 			}
 		}
+
+		// Rule 2: Non-comment arrives with stored comment — replay or discard
+		if let Some(comment) = self.pending_comment.take() {
+			if c != NO_WHITESPACE_BEFORE_KINDSET {
+				// Comment acts as separator — replay it
+				self.last_token = Some(comment.token());
+				self.sink.append(comment.compact());
```
keithamus: If you had `SourceCursor::EMTPY_COMMENT` it could stand in place of `comment.compact()` here.

kimjune01:
Pushed 12a9038c..70c966c1:

- L60 (whitespace coercion): collapsed to the shape you sketched, single guard checking `pending.is_none()`, `pending_comment.is_none()`, and `last_token != NO_WHITESPACE_AFTER_KINDSET`. Dropped `effective_last` and `last_cursor`. All 20 cursor_compact tests still pass.
- L74: subsumed by L60.
- L89 + L272: added `SourceCursor::EMPTY_COMMENT` backed by `Token::new_comment(CommentStyle::Block, 4)` (`pub(crate) const`). Replay path now appends `EMPTY_COMMENT` instead of `comment.compact()`, normalizing all comment content to `/**/`. Updated the @container test to feed `/*some comment text*/` and assert `/**/`.
- L91: removed.
- L247 / L255 / L270: already in 2393895f.

Deferred:

- L159 (trivia at csskit_transform and csskit callsites): out of scope here, filing as follow-up.

crates/css_parse/src/cursor_compact_write_sink.rs
kimjune01: Applied in 2393895f.

crates/css_parse/src/cursor_compact_write_sink.rs
kimjune01: Applied in 2393895f.

crates/css_parse/src/cursor_compact_write_sink.rs
kimjune01: Applied in 2393895f.

crates/css_parse/src/cursor_compact_write_sink.rs
kimjune01: The L60 simplification subsumes this. The pending/pending_comment/last_token guard replaces the last_cursor coercion. Existing 20 cursor_compact tests pass, including the `foo /**/bar` and `foo{/**/bar}` cases you called out.

crates/css_parse/src/cursor_compact_write_sink.rs
kimjune01: Applied in 70c966c1 via the EMPTY_COMMENT helper. @container test now feeds `/*some comment text*/` and asserts `/**/`.

crates/css_parse/src/cursor_compact_write_sink.rs
kimjune01: Filing as follow-up. Out of scope for this PR's blast radius (single-file, single-crate).

crates/css_parse/src/cursor_compact_write_sink.rs
kimjune01: Removed.

crates/css_parse/src/cursor_compact_write_sink.rs
kimjune01: Applied in 86f3f46e, exact shape you sketched. `effective_last` and `last_cursor` dropped.

crates/css_parse/src/cursor_compact_write_sink.rs
kimjune01: Added in 12a9038c. `Token::new_comment` is now `pub(crate) const`; replay path appends `EMPTY_COMMENT` instead of `comment.compact()`.

kimjune01:
@keithamus thanks for the careful reviews, and sorry that I wasted more of your time than it should have. 🙈

Full disclosure: this PR came from an autonomous pipeline ([kimjune01/sweep](https://github.com/kimjune01/sweep)). 

Looking back at the review pattern, you carried most of the implementation in `suggestion` blocks, which is polite but means we cost you more attention than self-implementing would have. CSS parsing has the kind of taste-density our pipeline doesn't reliably match, so we won't push further here. The branch is yours at this point. Merge, rework, or discard as you see fit.

If AI-generated PRs become a recurring load on csskit,  I can confidently recommend one of three filter options:

- [PR Quality Gate](https://github.com/kimjune01/sweep/blob/master/action.yml): judges PR quality not authorship; lets competent bots through, blocks lazy humans
- [anti-slop](https://github.com/peakoss/anti-slop): 34 configurable checks, built from maintaining Coolify (120 slop PRs/month)
- [agentscan-action](https://github.com/MatteoGabriele/agentscan-action): author-profile analysis, detects automation from account activity

Thanks again for engaging substantively.

keithamus:
> Full disclosure: this PR came from an autonomous pipeline ([kimjune01/sweep](https://github.com/kimjune01/sweep?rgh-link-date=2026-05-14T03%3A12%3A57Z)).

I was aware.

> Looking back at the review pattern, you carried most of the implementation in `suggestion` blocks, which is polite but means we cost you more attention than self-implementing would have. CSS parsing has the kind of taste-density our pipeline doesn't reliably match, so we won't push further here. The branch is yours at this point. Merge, rework, or discard as you see fit.

It was... mildly useful and entertaining to engage to see where this would go. If you want to spend LLM tokens fixing code in my projects, who am I to stop you.

> If AI-generated PRs become a recurring load on csskit, I can confidently recommend one of three filter options:

Thanks, I'll consider if it gets too burdensome.

PR #1113, opened 2026-05-13
csskit: Improve messaging for check commands with missing sheets

rlorenzo:
## Description

`NeverMatch::parse` was `unreachable!()`, relying on the invariant that `SelectorComponent::parse_selector_component` never dispatches to a placeholder type. That invariant doesn't hold for class selectors, pseudo-elements, functional pseudo-elements, and namespaced types in `QuerySelectorComponent`, so `csskit check` panicked on any real CSS that contained one of those constructs.

Two changes:

- `NeverMatch::parse` now returns `Diagnostic::unexpected` at the current cursor instead of panicking. The five affected dispatch paths (`Class`, `PseudoElement`, `LegacyPseudoElement`, `NsType`, `FunctionalPseudoElement`) now surface a normal parse error.
- `csskit check` now rejects an empty CSS file list and prints a hint when the first argument has a `.css` extension. That argument slot is the `.cks` rule sheet, so a `.css` file there is almost always a mistake, and was the exact path that produced the original bug report.

### 1. The user-facing repro (one argument)

```bash
echo '.foo { color: red; }' > /tmp/repro.css
npx csskit@latest check /tmp/repro.css
```

Before:

```
thread 'main' panicked at crates/csskit_ast/src/selector/query.rs:293:3:
internal error: entered unreachable code: NeverMatch should never be parsed
```

After:

```
error: no CSS files to check
usage: csskit check <rules.cks> <file1.css> [more.css...]
hint: `/tmp/repro.css` looks like a CSS file. The first argument is the csskit rule sheet (.cks).
Error: Parsing Failed!
```

This case is caught by the new CLI guard before the rule-sheet parser runs.

### 2. The underlying NeverMatch path

A `.cks` rule sheet that itself contains an unsupported selector also used to panic:

```bash
echo '.foo { color: red; }' > /tmp/rules.cks
echo 'a {}' > /tmp/dummy.css
csskit check /tmp/rules.cks /tmp/dummy.css
```

Before:

```
thread 'main' panicked at crates/csskit_ast/src/selector/query.rs:293:3:
internal error: entered unreachable code: NeverMatch should never be parsed
```

After: exits cleanly with no panic. (The Sheet parser currently swallows the resulting diagnostic via the `RuleVariants::parse_unknown_qualified_rule` recovery path, which is a separate pre-existing UX issue outside the scope of this fix; the panic itself is gone.) The new `assert_parse_error!` regression tests in `crates/csskit_ast/src/selector/query.rs` cover all five trigger constructs: `.foo`, `::before`, `::part(foo)`, `svg|*`, and `*|*`.

## Notes

This revises the previous version of this PR. The earlier revision fixed the panic by adding `Peek` bounds and peek-guards inside `css_parse`. Per maintainer feedback, `css_parse` is left untouched and the fix is now scoped to `csskit_ast` and the `csskit` CLI.

### DIFF
--- a/crates/csskit/src/color_ext.rs
+++ b/crates/csskit/src/color_ext.rs
@@ -57,6 +57,48 @@ where
 	format!("{}", text.on_truecolor(srgb.red, srgb.green, srgb.blue))
 }
 
+#[cfg(feature = "anstyle")]
+pub fn red<T: std::fmt::Display>(text: T) -> String {
+	use anstyle::{AnsiColor, Style};
+	let style = Style::new().fg_color(Some(anstyle::Color::Ansi(AnsiColor::Red)));
+	format!("{style}{text}{style:#}")
+}
+
+#[cfg(feature = "owo-colors")]
+#[cfg(not(feature = "anstyle"))]
+pub fn red<T: std::fmt::Display>(text: T) -> String {
+	use owo_colors::OwoColorize;
+	format!("{}", text.red())
+}
+
+#[cfg(feature = "anstyle")]
+pub fn bold_red<T: std::fmt::Display>(text: T) -> String {
+	use anstyle::{AnsiColor, Style};
+	let style = Style::new().bold().fg_color(Some(anstyle::Color::Ansi(AnsiColor::Red)));
+	format!("{style}{text}{style:#}")
+}
+
+#[cfg(feature = "owo-colors")]
+#[cfg(not(feature = "anstyle"))]
+pub fn bold_red<T: std::fmt::Display>(text: T) -> String {
+	use owo_colors::OwoColorize;
+	format!("{}", text.bold().red())
+}
+
+#[cfg(feature = "anstyle")]
+pub fn bold_green<T: std::fmt::Display>(text: T) -> String {
+	use anstyle::{AnsiColor, Style};
+	let style = Style::new().bold().fg_color(Some(anstyle::Color::Ansi(AnsiColor::Green)));
+	format!("{style}{text}{style:#}")
+}
+
+#[cfg(feature = "owo-colors")]
+#[cfg(not(feature = "anstyle"))]
+pub fn bold_green<T: std::fmt::Display>(text: T) -> String {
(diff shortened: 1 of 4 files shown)


keithamus (changes requested):
Thanks for tackling this one.

The SelectorComponent trait is supposed to handle all the peeking, so it can parse types with confidence. If we're going to add Peek requirements for these types I think it makes more sense to subsequently drop some of the peeking code in SelectorComponent. However I don't think that's what really wants to happen here.

The `NeverMatch` type is a bit of a hack for csskit's internal rule grammars, and the point is that we don't need things like class selectors for those. Given that I think `Parse` for NeverMatch can simply return a diagnostic failure instead of panicking. We should probably improve the `csskit check` command to have better messaging about parsing a csskit rule sheet vs a css style sheet, maybe even doing some magic to switch modes or something.

All this to say, I don't think this is the general solution we want. I think we want to avoid panics in csskit_ast, but maybe also put more into the `csskit` CLI, and probably not alter css_parse.

rlorenzo:
@keithamus The code has been updated. Is this what you had in mind?

repro/never-match-panic.css
```diff
@@ -0,0 +1,34 @@
+/* Reproducer for the `NeverMatch should never be parsed` panic in `csskit check`.
```
keithamus: Let's drop this entire file. It's never used. We have a `csskit-acceptance` task which would be used to test this behaviour instead.

keithamus:
> @keithamus The code has been updated. Is this what you had in mind?

Yes I think this is looking a lot closer to what I had imagined.

repro/never-match-panic.css
rlorenzo: Done in 1a1500b0: dropped `repro/never-match-panic.css`. The unit tests in `crates/csskit_ast/src/selector/query.rs` already cover the five NeverMatch variants, so no acceptance case added. Happy to add one if you want end-to-end coverage too.

keithamus (approved):
Thanks for this one @rlorenzo!

PR #1114, opened 2026-05-13
fix(css_parse): preserve whitespace around +/- in calc()

rlorenzo:
## Description

- Spec requires whitespace around `+`/`-` in `<calc-sum>`; WebKit treats unspaced operators as invalid and falls back to the property's initial value
- The compact sink dropped that whitespace whenever the prior token was in the no-whitespace-after set (e.g. a closing paren), emitting `calc(var(--x)- 1px)` instead of `calc(var(--x) - 1px)`

## Example

Input:

```css
.a { left: calc(var(--col-gap) / 2 + var(--date-col)); }
.b { right: calc((var(--col-gap) / -2) - 5px); }
```

`csskit min` output before this PR (invalid in WebKit; both rules fall back to their initial values):

```css
.a{left:calc(var(--col-gap)/ 2+ var(--date-col))}
.b{right:calc((var(--col-gap)/ -2)- 5px)}
```

After this PR (whitespace preserved around `+`/`-`):

```css
.a{left:calc(var(--col-gap)/ 2 + var(--date-col))}
.b{right:calc((var(--col-gap)/ -2) - 5px)}
```

Spec reference: [CSS Values and Units Level 4 §10.1](https://www.w3.org/TR/css-values-4/#calc-syntax). *"The `+` and `-` operators must be surrounded by whitespace."*

## Notes


crates/css_parse/src/cursor_compact_write_sink.rs
```diff
@@ -31,10 +33,28 @@ impl<'a, T: SourceCursorSink<'a>> CursorCompactWriteSink<'a, T> {
...
+			let prev_is_whitespace = prev == Kind::Whitespace;
+			let whitespace_required = prev_is_whitespace && {
+				let last_is_calc_sum_op = self.last_token.is_some_and(|t| {
+					matches!(t.char(), Some('+' | '-')) && t == AssociatedWhitespaceRules::EnforceAfter
+				});
+				last_is_calc_sum_op || {
+					let next = c.token();
+					matches!(next.char(), Some('+' | '-')) && next == AssociatedWhitespaceRules::EnforceBefore
+				}
+			};
```
keithamus: This isn't quite right. We don't need to check for the exact characters here; `CompactCursorWriteSink` should be agnostic to the domain specifics of the host language as it's for parsing CSS like languages, not just CSS. This is exactly why `AssociatedWhitespaceRules` exists. 

This code can be simplified by just checking for a token's `AssociatedWhitespaceRules` to check that - if enforced - whitespace should be emitted.

keithamus (changes requested):
Thanks for working in this, but this doesn't feel quite right. There's some overly specific checks and it's doing redundant work.

The _intent_ behind calc requiring such conditions is that a value such as `2px + 2px` is unambiguously 3 tokens: `<dimension> <delim> <dimension>`. Eliding whitespace between the delim and dimension results in a new token: `+2px`, so it becomes `<dimension> <dimension>`. Worse still that might lead to the `+` being elided so the resulting code would be `2px 2px`. Likewise `2px - 2px` (`<dimension> <delim> <dimension>`) would be compacted and made into `2px-2px` and `px-2px` is a valid ident meaning that would produce a single `<dimension>` token. It's for this reason that `-` and `+` need associated whitespace rules.

Some of this work is already done in `needs_separator_for`. A `<delim '-'> <number|dimension>` or `<delim '-'> <number|dimension>` already gets whitespace enforcement, meaning some of the code here is already redundant - we don't need to check that. Likewise `<number|dimension> <delim '-'>` cases are also covered. You can see this today in the playground: https://csskit.rs/playground/?code=S8vPr05KLLJKTsxJ1jAsqFDQVgCSmrUA&format=minify, https://csskit.rs/playground/?code=S8vPr05KLLJKTsxJ1jAsqFDQVQCSmrUA&format=minify.

Where this starts to fall apart and what your tests highlight is that `needs_separator_for` doesn't cover cases like `<delim ')'> <delim '+'>`; and it shouldn't. `ComponentValues` maybe could, but also really the issue goes deeper:

We don't currently properly parse Calc expressions. Right now they're parsed as component values, and not a proper Calc tree. The real work here is to build parsing for calc expressions and setting the correct `AssociatedWhitespaceRules` on the operators within calc expressions to ensure they have a space before and after them.

We might be able to get by with some subtle changes to ComponentValues in the meantime, which _might_ need CompactCursorWriteSink changes to add the `AssociatedWhitespaceRules` checks, but we certainly don't need to check individual characters in CursorCompactWriteSink.

Hope this helps you (or your AI) come up with a more appropriate solution here.

rlorenzo:
@keithamus I believe the fix in https://github.com/csskit/csskit/pull/1114/commits/6239fab4726b7db53691c64b34ccee61401329de is what you were mentioning.

But for the Calc AST, the longer-term fix, the changes are much more extensive, and I am not sure if you would want to accept them. It is available in a PR view in my forked repo if it is of interest to you: https://github.com/rlorenzo/csskit/pull/1/changes

crates/css_parse/src/syntax/component_values.rs
```diff
@@ -34,8 +34,13 @@ impl<'a> Parse<'a> for ComponentValues<'a> {
 				break;
 			}
 			if let Some(mut value) = p.parse_if_peek::<ComponentValue>()? {
+				// Only flag `EnforceBefore` on Delims that are also flanked by whitespace on the
+				// trailing side (EnforceAfter set by `ComponentValue::parse`). Single-flanked
+				// Delims like `.` in `} .foo{}` shouldn't force whitespace preservation in the
+				// sink.
```
keithamus: This comment isn't really necessary, especially as it only describes the new condition, not the pre-existing ones.

```suggestion
```

crates/css_parse/src/cursor_compact_write_sink.rs
```diff
@@ -31,10 +33,22 @@ impl<'a, T: SourceCursorSink<'a>> CursorCompactWriteSink<'a, T> {
 			self.pending = None;
 			let is_redundant_semi = prev == Kind::Semicolon
 				&& (c == REDUNDANT_SEMI_KINDSET || self.last_token.is_some_and(|c| c == REDUNDANT_SEMI_KINDSET));
-			let no_whitespace_after_last =
-				prev == Kind::Whitespace && self.last_token.is_some_and(|c| c == NO_WHITESPACE_AFTER_KINDSET);
+			// Honor `AssociatedWhitespaceRules` set by upstream parsers: if the last token
+			// requires whitespace after it or the next token requires whitespace before it,
+			// the pending whitespace must be preserved even when the surrounding token kinds
+			// would otherwise allow trimming. This is how spec-required whitespace (e.g.
+			// around `+`/`-` in `calc()`) survives minification.
```
keithamus: Same here. 

```suggestion
```

crates/css_parse/src/cursor_compact_write_sink.rs
```diff
@@ -164,6 +178,22 @@ mod test {
...
+	#[test]
+	fn test_preserves_whitespace_around_calc_sum_operators() {
+		// The CSS spec requires whitespace around `+` and `-` in calc() <calc-sum>; without
+		// it, WebKit treats the expression as invalid. See:
+		// https://www.w3.org/TR/css-values-4/#calc-syntax
+		//
+		// The sink honors `AssociatedWhitespaceRules` generically: any Delim that was flanked
+		// by whitespace on both sides in the source keeps both surrounding spaces. This
+		// preserves the spec-critical spacing around `+`/`-` and, as a side effect, around
+		// `*`/`/` too (slightly less aggressive minification, but still valid CSS).
```
keithamus: This comment is also redundant, but not for the same reasons. This one:

- Mentions only WebKit for reasons I'm not fully sure why (I guess an AI prompt?)
- Descibes sink behaviour rather than what the test is testing.
- Describes potentially temporary behavior that may end up descrynchronising with the test.


```suggestion
```

keithamus (approved):

PR #1115, opened 2026-05-14
css_lexer: Keep separator between string and value tokens

rlorenzo:
## Description

The compact serializer dropped whitespace between strings and following ident/number tokens, so

    font-variation-settings: "opsz" 14

was minified to

    font-variation-settings:"opsz"14

WebKit rejects the latter and falls back to `normal`, breaking optical-sizing on variable fonts. font-feature-settings has the same `<string> <ident|integer>` grammar and was affected too. Extend `needs_separator_for` so a string followed by ident, function, url, bad-url, number, or dimension preserves a space.

## Notes

### DIFF
--- a/crates/css_ast/src/functions/image_set_function.rs
+++ b/crates/css_ast/src/functions/image_set_function.rs
@@ -45,7 +45,7 @@ mod tests {
 
 	#[test]
 	fn test_writes() {
-		assert_parse!(CssAtomSet::ATOMS, ImageSetFunction, "image-set('image.jpg'1x,'image.jpg'2x)");
+		assert_parse!(CssAtomSet::ATOMS, ImageSetFunction, "image-set('image.jpg' 1x,'image.jpg' 2x)");
 		assert_parse!(
 			CssAtomSet::ATOMS,
 			ImageSetFunction,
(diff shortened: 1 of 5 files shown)


keithamus:
@rlorenzo do you have a reduced test case demonstrating this bug in WebKit? I had a look at the source to try and see if I could find the bug and I couldn't. I then knocked up a test page:

```html
<!DOCTYPE html>
<style>
@import url('https://fonts.googleapis.com/css2?family=Inter:opsz,wght@14..32,100..900&display=swap');
body { font-family: Inter, sans-serif; font-size: 48px; margin: 20px; }
.overlay-container {
  position: relative;
  height: 48px;
}
.baseline, .test {
  position: absolute;
  top: 0;
  left: 0;
}
.baseline {
  font-variation-settings: normal;
  color: red;
  z-index: 1;
}
.test {
  color: blue;
  z-index: 2;
  mix-blend-mode: difference;
}
.spaced .test { font-variation-settings: "opsz" 14; }
.unspaced .test { font-variation-settings:"opsz"14; }
</style>

<div class="overlay-container spaced">
  <div class="baseline">Waffles Typography</div>
  <div class="test">Waffles Typography</div>
</div>
<div class="overlay-container unspaced">
  <div class="baseline">Waffles Typography</div>
  <div class="test">Waffles Typography</div>
</div>
```

Here's Chrome:
image

Firefox:
image

Safari:
image

I think the issue that you maybe saw was that WebKit seems to be an outlier when it comes to applying the font-variation-settings? They _are_ parsed and they _are_ applied, and spacing doesn't seem to be the issue. All browsers parse both space and unspaced fine.

rlorenzo:
@keithamus You are right. Closing PR.

